Issue #25, opened 2022-12-16
IngestArch: Add structure to  architectures topics


karenzone:
Apply consistent intro structure for consistency across architectures
* [x] #26
* [x]  #27
* [x] #31
* [x] #32
* [x] #33

karenzone:
Note to self @karenzone: https://docs_2578.docs-preview.app.elstc.co/guide/en/ingest/current/index.html

PR #32, opened 2022-12-28
IngestArch: Kafka as message queue

karenzone:
* [x] Break out Kafka as message queue (6): 
  * [x] 6a - Using LS Elastic Agent input and ES Sink
  * [x] 6b - Using LS to Kafka to LS to ES
* [x] Other corrections and content tweaks

Related: #25


karenzone (commented):
Builds cleanly and renders as expected

PR #33, opened 2022-12-29
IngestArch: Add more content and structure to architectures

karenzone:
Pulled in top level topic for 2-Custom integrations
Made corrections, and more content and structure

Related: #25


karenzone (commented):
Builds cleanly and renders locally as expected

Issue #36, opened 2023-01-11
IngestArch: Restructure content by architecture and nest use cases


karenzone:
Current structure is organized by use case for P1 items. 
Restructure content for each architecture and nest use cases for each. 

#### ToDo
Content analysis to figure out where content and diagrams fit best per architecture.  For some, the diagram and most of descriptive content makes sense at the top level with details for each flavor underneath. 
As the architectures and use cases become more complex, we'll require more differentiation and details pushed lower into the structure.  (The various use cases for Agent -> LS -> ES, for example.)

PR #37, opened 2023-01-12
IngestArch: Restructure source files and include statements

karenzone:
Related: #36


karenzone (commented):
Builds cleanly locally and renders as expected

PR #38, opened 2023-01-13
IngestArch: Restructure high-level archs to pull in flavors

karenzone:
Related: #36


karenzone (commented):
Builds cleanly locally and renders as expected.

PR #40, opened 2023-01-18
IngestArch: Restructure to bring ea-proxy to top level

karenzone:
Related: #36


karenzone (commented):
Builds cleanly locally and renders as expected

Issue #42, opened 2023-01-26
IngestArch: Create and add architecture graphics for Phase 1 architectures


karenzone:
Add high-level graphics for each architecture/use case

- [ ] 1 - EA to ES
   - [ ] 1a - EA installed
   - [ ] 1b APIs for collection
- [ ]  15 - EA to proxy to ES
- [ ]  EA to Logstash to ES
   - [ ] 10/11 - EA to Logstash (for enrichment) to ES
   - [ ] 12 - EA to Logstash (PQ for buffering) to ES
   - [ ] 14 - EA to Logstash (to bridge networks) to ES
   - [ ] 13 - EA to Logstash to multiple destinations
- [ ] 16 - EA to ES with Kafka as middleware message queue
   - [ ] 16a - Kafka to ES Sink
   - [ ] 16b - Kafka to Logstash
- [ ] Logstash to ES

amitkanfer:
@karenzone I'm moving this to sprint 8 (7 is over)

karenzone:
Completed in Sprint 7

PR #45, opened 2023-01-27
IngestArch: Add and include graphics

karenzone:
Related: #42

Adds and includes graphics for: 
* LS to ES
* EA to LS to ES (Same image used for overview and PQ. May break out into separate images.)
* Agent with Kafka as middleware (ES sink)


docs/en/ingest-arch/1x-agent-ls.asciidoc
```diff
@@ -1,9 +1,10 @@
 [[agent-ls]]
 == {agent} to {ls} to {es}
 
-Ingest model::
+Ingest models::
 {agent} to {es} using {ls} for advanced use cases such as enrichment, buffering, network bridging, multiple endpoints
 
+image::images/ea-lspq-es.png[Image showing {agent} to {ls} to {es}]
```
karenzone, line 7: Currently reusing PQ image, but will replace with more generic image

karenzone:
@karenzone: DB#34882

PR #48, opened 2023-01-31
IngestArch: Add and include more graphics

karenzone:
Related: #42 

Adds graphics and includes for:
* EA to proxy to ES
* EA to LS (as bridge)
* EA to LS (for enrichment)


karenzone:
@karenzone: DB #35071

Issue #47, opened 2023-01-27
Minor corrections for doc/graphics


karenzone:
I'm making some corrections as I go. Adding a list here for thing I (and others) spot so that we can address them. 

### Punch list for doc/graphics

Graphics:
* [ ] ea-apis (agent-apis.html) - Label components in External Infrastructure box
* [ ] ea-proxy-es (agent-proxy) - Redraw/replace graphic to follow graphic style used for other topics. 
* [ ] ea-kafka (agent-kafka-es) - Replace dotted line between EA and Kafka with a Logstash instance (because it's common to both flavors)
* [ ] ea-kafka-essink (agent-kafka-essink)
    * [ ] Align along top. 
    * [ ] Refine ES Sink component box
* [ ] ls-to-es (ls-for-input.html) - Align along top

amitkanfer:
@karenzone I'm moving this to sprint 8

karenzone:
Completed

PR #53, opened 2023-02-10
IngestArch: Update graphics to show control path

karenzone:
Related #47

* [ ] [EA to ES: APIs for collection](https://www.elastic.co/guide/en/ingest/current/agent-apis.html) (ea-apis.png)
    @mukeshelastic, please confirm that this is the change you want to see before I update the remaining diagrams.


mukeshelastic:
quick feedback- there is no proxy here.. replace proxy with fleet, APIs label is between ext infra and elastic agent..  also add this text somewhere to describe how this is working.. Elastic agent is running on a user managed host, that is collecting data about the ext infra via API..

karenzone:
Thanks for the quick response, @mukeshelastic .

> quick feedback- there is no proxy here.. replace proxy with fleet, 

Will do. 

> APIs label is between ext infra and elastic agent

My bad. I knew that.  Will fix. 

> also add this text somewhere to describe how this is working.. Elastic agent is running on a user managed host, that is collecting data about the ext infra via API..

We've been putting control path descriptions in section content rather than in graphics.  I'll make that change for the next round so that you can see it.

karenzone:
@mukeshelastic, I've incorporated your feedback in the latest commit.  
Preview available here: https://ingest-docs_53.docs-preview.app.elstc.co/guide/en/ingest/master/agent-apis.html

Please let me know if we're on the right path.  If so, I'll implement these changes (as appropriate) for other architectures.

karenzone:
@mukeshelastic Please take a look at the updated graphic and let me know if it's what you wanted.  When I get confirmation (or request for changes), I'll keep this work moving forward.

mukeshelastic:
The diagram looks right. let's replicate this diagram in other archs.

docs/en/ingest-arch/15-proxy.asciidoc
```diff
@@ -6,8 +6,8 @@ image::images/ea-proxy-fs-es.png[Image showing connections between {agent} and {
 image::images/ea-fs-proxy-es.png[Image showing connections between {agent} and {es} using a proxy]
 
 Ingest model::
-Data path: {agent} to proxy to {es} +
-Control path: 
+Control path: {agent} to proxy to {es} +
+Data path: 
 * {agent} to proxy to {fleet-server} to {es} (for {fleet-server} on {ecloud}) or +
```
karenzone, line 11: Correct data path and control path in text

docs/en/ingest-arch/1b-apis.asciidoc
```diff
@@ -4,7 +4,8 @@
 image::images/ea-apis.png[Image showing {agent} collecting data using APIs and sending to {es}]
 
 Ingest model::
-{agent} to {es} using APIs for collection
```
karenzone, line 7: Add control path per review comment

PR #63, opened 2023-03-23
IngestArch: Add introductory table for architectures

karenzone:
Related: #61

PR #66, opened 2023-03-29
Revert "IngestArch: Add introductory table for architectures"

flexitrev:
Reverts elastic/ingest-docs#63


karenzone (commented):
We'll let it stand as merged. :-)

Issue #70, opened 2023-03-31
IngestArch: Flesh out LS use cases


karenzone:
Comments moved over from #62 

#### [Elastic Agent to Logstash (for enrichment) to Elasticsearch](https://www.elastic.co/guide/en/ingest/current/ls-enrich.html)
* [x]  On the enrichment [doc](https://www.elastic.co/guide/en/ingest/current/ls-enrich.html), i would link to a few enrich processors we have (there's a link at the bottom to "Transform" which is a bit different than just "enrich") (#74)
* [x]  "Data sources that change frequently, such as stock ticker data" - why is this an example to use Logstash? does it imply that Agent is not fast enough? or is it because the enriched data is "public" data source and thus we need Logstash's custom HTTP/tcp input plugin? (#77)
* [x] in [this](https://www.elastic.co/guide/en/ingest/current/ls-enrich.html#ls-enrich-proc) process overview, worth adding how to configure logstash as an output to Agent? (in the agent policy UI)
* [x] Same thing for all 4 Agent to LS documents 👍 
  
##### [Elastic Agent to Logstash to Elasticsearch: Logstash to bridge networks](https://www.elastic.co/guide/en/ingest/current/ls-networkbridge.html#ls-networkbridge)
* [x] Worth adding a visual in the [diagram](https://www.elastic.co/guide/en/ingest/current/ls-networkbridge.html) to show the two different networks that LS is used to bridge between. (#74)
* [ ] "Use when"... the same explanation was given to put a proxy between agents and ES, so what makes this use case unique to use Logstash?
* [x] Examples are missing...
     From @karenzone: I added some information (in #77) so that this section is no longer empty. We can add more details or specific examples if needed.

karenzone:
@roaksoax @flexitrev @jsvd (and other interested parties):  We need your input, please.

PR #74, opened 2023-04-04
IngestArch: Update Logstash-as-bridge graphic

karenzone:
Related: #70

Responding to review comment for [Elastic Agent to Logstash to Elasticsearch: Logstash to bridge networks](https://www.elastic.co/guide/en/ingest/current/ls-networkbridge.html#ls-networkbridge)
* [ ] Worth adding a visual in the [diagram](https://www.elastic.co/guide/en/ingest/current/ls-networkbridge.html) to show the two different networks that LS is used to bridge between.


docs/en/ingest-arch/images/ea-ls-bridge.png
flexitrev, line 1: Image looks good, showing two different clusters of agents helps illustrate the use case.

flexitrev (approved):
Updated image is a good illustration.

PR #96, opened 2023-04-12
Add note on location of proxy settings for artifact download

bmorelli25:
This PR replaces https://github.com/elastic/observability-docs/pull/2844.

### DIFF
--- a/docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
+++ b/docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
@@ -27,7 +27,7 @@ Configure proxy settings for {agent} when it must connect through a proxy server
 to:
 
 * Download artifacts from `artifacts.elastic.co` for subprocesses or binary
-upgrades
+upgrades (use <<fleet-agent-binary-download-settings>>)
 * Send data to {es}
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/fleet-agent-proxy-support.asciidoc
```diff
@@ -132,6 +132,8 @@ HTTP_PROXY=http://my.proxy:8080
 
 After adding environment variables, restart the service.
 
+NOTE: If you use a proxy server to download artifacts from `artifacts.elastic.co`, configure <<fleet-agent-binary-download-settings>>.
```
karenzone, line 135:
```suggestion
NOTE: If you use a proxy server to download new agent versions from `artifacts.elastic.co` for upgrading, configure <<fleet-agent-binary-download-settings>>. 
```

karenzone:
@elasticmachine run elasticsearch-ci/docs

AndersonQ:
@TheRiffRafi you're right, right now there is no way to configure a proxy specifically for downloading artifacts, the current solution is, as you said, a redirect at the proxy level.

>the value could be something like `myproxy.net:1234/artifacts/downloads` and that in the proxy redirects to `artifacts.elastic.co/downloads`.

PR #166, opened 2023-04-24
Fix Fleet Server pull and issue attributes

kilfoyle:
These attributes are currently pointing to the elastic/beats repo, which I assume was a copy & paste error.

I noticed this in the 7.17 branch (where it breaks the Release Notes link to a PR) so this will need to be backported to all available branches.


kilfoyle:
Closing. This doesn't look right.

Issue #101, opened 2022-11-22
[REQUEST]: document Elastic Agent HTTPS Proxy settings and configuration


jlind23:
### Description

the use case is that the Agents deployed in the fleet, all have their own individual certificates. They are communicating to Fleet Server and Elasticsearch via a HTTPS proxy. Customer would provide a Certificate Authority for the certificates they use on the Agents.

We need to document for the user as to how this can be done.

### Resources

https://github.com/elastic/ingest-dev/issues/1274#issuecomment-1316010045

Michel's comment already described all the config and how they should be set up.

### Collaboration

The product team will provide the initial content.

### Point of contact.

**Main contact:** @nimarezainia @michel-laterman 

**Stakeholders:**

nimarezainia:
hi @kilfoyle, sorry for the delay. I have the content rewrite placed here: https://docs.google.com/document/d/19UyFdeB9uLDasfqZiRxnGK0SfssiOeKwvgGisCeovZ0/edit?usp=sharing

I have commented in RED what I suggest should change. Could we comment via the doc to finalize or happy to have a call about it. Also happy to just review the final copy. lmk thanks.

kilfoyle:
@nimarezainia Thanks for the excellent rewrite! Here's a [PR](https://github.com/elastic/ingest-docs/pull/271) with the changes.

PR #169, opened 2023-04-24
Add Elastic Agent proxy config docs

kilfoyle:
This is just an early draft.


kilfoyle:
@karenzone  Certainly. The parent issue is here: https://github.com/elastic/ingest-docs/issues/101

PR #178, opened 2023-04-28
Update commands.asciidoc

insukcho:
Adding more specific information to avoid confusion when the customer has Root CA + Intermediate CA. It could give more clear format (PEM encoded) and always add the full certificate chain.


insukcho:
I'm adding this to https://github.com/elastic/ingest-docs/pull/179

PR #181, opened 2023-04-28
Change the known issue number to pint to the correct one

AndersonQ:
The known issue that "may prevent upgrades scheduled to execute at a later time from running" is pointing to the wrong GH issue.

The original PR https://github.com/elastic/observability-docs/pull/2726 correctly states:
> Add https://github.com/elastic/elastic-agent/issues/2343 as a known issue for 8.6.x releases

but  https://github.com/elastic/elastic-agent/issues/2342 was used instead.

This PR fixes that.


pierrehilbert (approved):
Good catch!

PR #195, opened 2023-05-04
Add known issue for agent 8.6.2 about not migrating unencrypted config

pchila:
Create a new entry for a known issue of agent not loading fleet configuration when upgrading from a version using unencrypted config `fleet.yml` to encrypted one `fleet.enc`

### DIFF
--- a/docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
+++ b/docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
@@ -117,11 +117,11 @@ Settings with input blocks, such as Max agents are still effective.
 ====
 
 [[known-issue-issue-2249-8.6.2]]
-.{agent} does not load {fleet} configuration when upgrading to {agent} version 8.6.x from version 8.2 or earlier.
+.{agent} does not load {fleet} configuration when upgrading to {agent} version 8.6.x from version 8.2.x or earlier.
 [%collapsible]
 ====
 *Details* +
-{agent} will not load `fleet.yml` information after upgrading to {agent} version 8.6.x from version 8.2 or earlier
+{agent} will not load `fleet.yml` information after upgrading to {agent} version 8.6.x from version 8.2.x or earlier
 This is due to the unencrypted config stored in `fleet.yml` not being migrated correctly to the encrypted config store `fleet.enc`.
 For a managed agent the symptom manifests as the inability to connect to {fleet} after upgrade with the following log:
 
@@ -326,11 +326,11 @@ Settings with input blocks, such as Max agents are still effective.
 ====
 
 [[known-issue-issue-2249-8.6.1]]
-.{agent} does not load {fleet} configuration when upgrading to {agent} version 8.6.x from version 8.2 or earlier.
+.{agent} does not load {fleet} configuration when upgrading to {agent} version 8.6.x from version 8.2.x or earlier.
 [%collapsible]
 ====
 *Details* +
-{agent} will not load `fleet.yml` information after upgrading to {agent} version 8.6.x from version 8.2 or earlier
+{agent} will not load `fleet.yml` information after upgrading to {agent} version 8.6.x from version 8.2.x or earlier
 This is due to the unencrypted config stored in `fleet.yml` not being migrated correctly to the encrypted config store `fleet.enc`.
 For a managed agent the symptom manifests as the inability to connect to {fleet} after upgrade with the following log:
 
@@ -505,11 +505,11 @@ Settings with input blocks, such as Max agents are still effective.
 ====
 
 [[known-issue-issue-2249-8.6.0]]
-.{agent} does not load {fleet} configuration when upgrading to {agent} version 8.6.x from version 8.2 or earlier.
+.{agent} does not load {fleet} configuration when upgrading to {agent} version 8.6.x from version 8.2.x or earlier.
 [%collapsible]
 ====
 *Details* +
-{agent} will not load `fleet.yml` information after upgrading to {agent} version 8.6.x from version 8.2 or earlier
+{agent} will not load `fleet.yml` information after upgrading to {agent} version 8.6.x from version 8.2.x or earlier
 This is due to the unencrypted config stored in `fleet.yml` not being migrated correctly to the encrypted config store `fleet.enc`.
 For a managed agent the symptom manifests as the inability to connect to {fleet} after upgrade with the following log:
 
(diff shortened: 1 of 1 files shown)


belimawr (approved):

docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
```diff
@@ -116,6 +116,25 @@ Custom yaml settings are silently ignored by {fleet}.
 Settings with input blocks, such as Max agents are still effective.
 ====
 
+[[known-issue-issue-2249-8.6.2]]
+.{agent} does not load fleet configuration after upgrade from unencrypted config
```
cmacknz, line 120: You need to specify which versions have unencrypted configuration. If you are a user how do you know this applies to you?

cmacknz: Save them from having to read the full summary, which is collapsed by default

docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
```diff
@@ -116,6 +116,25 @@ Custom yaml settings are silently ignored by {fleet}.
...
+[%collapsible]
+====
+*Details* +
+{agent} will not load `fleet.yml` information after upgrading to a version that uses encrypted configuration (> 8.3).
+For a managed agent the symptom manifests as the inability to connect back to fleet after upgrade and the following log:
```
cmacknz, line 125:
```suggestion
For a managed agent the symptom manifests as the inability to connect to {fleet} after upgrade with the following log:
```

docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
cmacknz, line 120:
```suggestion
.{agent} does not load {fleet} configuration after upgrade from unencrypted config
```

docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
```diff
@@ -116,6 +116,25 @@ Custom yaml settings are silently ignored by {fleet}.
...
+[source,shell]
+----
+Error: fleet configuration is invalid: empty access token
+...
+----
+For more information refer to {agent-issue}2249[#2249].
+
+*Impact* +
+{agent} loses the ability to connect back to {fleet} after upgrade.
+Fix for this issue is available in version 8.7.0 and higher, for versions 8.3-8.6 a re-enroll is necessary for the agent to connect back to {fleet}
```
cmacknz, line 136: Does this actually affect 8.3-8.6? I thought it only affected 8.6?

pchila: I thought I had tested 8.3 and 8.4 in the original issue but I actually didn't, I only tested 8.2 -> 8.6
I will test 8.2 -> 8.3, 8.2 -> 8.4, 8.2 ->8.5 upgrade scenario to confirm

pchila: 8.2.0 -> 8.3.3, 8.2.0 ->8.4.3, 8.2.0 -> 8.5.3 all work correctly I will update the description to reflect that

docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
```diff
@@ -116,6 +116,26 @@ Custom yaml settings are silently ignored by {fleet}.
 Settings with input blocks, such as Max agents are still effective.
 ====
 
+[[known-issue-issue-2249-8.6.2]]
+.{agent} does not load {fleet} configuration after upgrading from {agent} version 8.2 or earlier
```
cmacknz, line 120:
```suggestion
.{agent} does not load {fleet} configuration when upgrading to {agent} version 8.6.x from version 8.2 or earlier
```

Based on https://github.com/elastic/ingest-docs/pull/195#discussion_r1185267796 the affected versions still need to be clearer. I want to be extra clear that this only affects upgrades to 8.6.x, even though these are the 8.6 release notes.

docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
```diff
@@ -116,6 +116,26 @@ Custom yaml settings are silently ignored by {fleet}.
...
+[%collapsible]
+====
+*Details* +
+{agent} will not load `fleet.yml` information after upgrading from a version up to 8.2 to 8.6.
```
cmacknz, line 124:
```suggestion
{agent} will not load `fleet.yml` information after upgrading to {agent} version 8.6.x from version 8.2 or earlier
```

Let's just use the same wording for versions from above.

docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
```diff
@@ -116,6 +116,26 @@ Custom yaml settings are silently ignored by {fleet}.
 Settings with input blocks, such as Max agents are still effective.
 ====
 
+[[known-issue-issue-2249-8.6.2]]
```
cmacknz, line 119: This can be copied to the 8.6.1 and 8.6.0 release notes which are in this same document. This is a bit annoying but let's make sure users are likely to read this regardless of the version of 8.6 they are targeting.

* 8.6.1 notes https://github.com/elastic/ingest-docs/blob/main/docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc#release-notes-8.6.1
* 8.6.0 notes https://github.com/elastic/ingest-docs/blob/main/docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc#release-notes-8.6.0

docs/en/ingest-management/release-notes/release-notes-8.6.asciidoc
```diff
@@ -116,6 +116,26 @@ Custom yaml settings are silently ignored by {fleet}.
...
+.{agent} does not load {fleet} configuration when upgrading to {agent} version 8.6.x from version 8.2 or earlier
```
cmacknz, line 120:
```suggestion
.{agent} does not load {fleet} configuration when upgrading to {agent} version 8.6.x from version 8.2.x or earlier
```

For consistency we should use 8.2.x to ensure people assume this applies to all 8.2.x version. I'll just make this comment one time but it applies in multiple places.

cmacknz (approved):
Thank you!

Issue #240, opened 2023-05-31
We should document this behaviour in our public doc page for standalone agent


jlind23:
We should publicly document that Elastic Agent file paths can not exceed 103 or else the path will be rewrite to something like:
`/tmp/elastic_agent/$sha256(path).sock` instead of `unix:///usr/share/elastic-agent/state/data/tmp/kubernetes-metrics-60f88f50-c873-11ed-9baf-09fb5640c56a.sock` where $sha256(path) is the SHA256 hash of the original path that is too long.

jlind23:
@kilfoyle @karenzone This should be fairly small to document, would it be feasible to add it to our 8.8 docs as soon as possible?

kilfoyle:
@jlind23  I'll add it in today.

PR #241, opened 2023-05-31
Add file path length restriction for standalone agent

kilfoyle:
This PR:
 - Adds a note to the [Install standalone Elastic Agents](https://www.elastic.co/guide/en/fleet/current/install-standalone-elastic-agent.html) page indicating that the install path can't exceed 103 characters.
- Combines this with the other two notes at the top of that page into a single "restrictions" block.

Preview:
Screenshot 2023-05-31 at 1 23 29 PM

Closes: https://github.com/elastic/ingest-docs/issues/240

### DIFF
--- a/docs/en/ingest-management/elastic-agent/install-standalone-elastic-agent.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/install-standalone-elastic-agent.asciidoc
@@ -21,7 +21,7 @@ Note the following restrictions for running standalone {agent}:
 versions automatically. When you upgrade the integration in {kib}, you'll
 need to update the standalone policy manually.
 
-* The file path where {agent} is installed cannot exceed 103 characters.
+* {agent} file paths cannot exceed 103 characters. This restriction applies to all paths, including the {agent} install directory and the data path used by any integrations.
 
 * You can install only a single {agent} per host.
 ====
--- a/docs/en/ingest-management/elastic-agent/install-standalone-elastic-agent.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/install-standalone-elastic-agent.asciidoc
@@ -13,11 +13,18 @@ We recommend using <<install-fleet-managed-elastic-agent,{fleet}-managed {agent}
 when possible, because it makes the management and upgrade of your agents
 considerably easier. 
 
-IMPORTANT: Standalone agents are unable to upgrade to new integration package
+[IMPORTANT]
+====
+Note the following restrictions for running standalone {agent}:
+
+* Standalone agents are unable to upgrade to new integration package
 versions automatically. When you upgrade the integration in {kib}, you'll
 need to update the standalone policy manually.
 
-NOTE: You can install only a single {agent} per host.
+* The file path where {agent} is installed cannot exceed 103 characters.
+
+* You can install only a single {agent} per host.
+====
 
 {agent} can monitor the host where it's deployed, and it can collect and forward
 data from remote services and hardware where direct deployment is not possible.


docs/en/ingest-management/elastic-agent/install-standalone-elastic-agent.asciidoc
```diff
@@ -13,11 +13,18 @@ We recommend using <<install-fleet-managed-elastic-agent,{fleet}-managed {agent}
...
+[IMPORTANT]
+====
+Note the following restrictions for running standalone {agent}:
+
+* Standalone agents are unable to upgrade to new integration package
 versions automatically. When you upgrade the integration in {kib}, you'll
 need to update the standalone policy manually.
 
-NOTE: You can install only a single {agent} per host.
+* The file path where {agent} is installed cannot exceed 103 characters.
```
jlind23, line 24: This is not just about the installation path, all paths, including data path for specific integration shouldn't exceed 103 characters.
Maybe @joshdover or @cmacknz have a better wording here.

kilfoyle: How about this:
> Elastic Agent file paths cannot exceed 103 characters. This restriction applies to all paths, including the Elastic Agent install directory and the data path used by any integrations.

kilfoyle:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.8|[](https://github.com/elastic/ingest-docs/pull/243)|

Note: Successful backport PRs will be merged automatically after passing CI.

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport)

docs/en/ingest-management/elastic-agent/install-standalone-elastic-agent.asciidoc
```diff
@@ -13,11 +13,18 @@ We recommend using <<install-fleet-managed-elastic-agent,{fleet}-managed {agent}
...
+[IMPORTANT]
+====
+Note the following restrictions for running standalone {agent}:
+
+* Standalone agents are unable to upgrade to new integration package
 versions automatically. When you upgrade the integration in {kib}, you'll
 need to update the standalone policy manually.
 
-NOTE: You can install only a single {agent} per host.
+* {agent} file paths cannot exceed 103 characters. This restriction applies to all paths, including the {agent} install directory and the data path used by any integrations.
```
cmacknz, line 24: > * {agent} file paths cannot exceed 103 characters. This restriction applies to all paths, including the {agent} install directory and the data path used by any integrations.

I don't think this is true in general, this limitation only applies to paths that represent Unix sockets. The case that originally hit this actually constructs a Unix file path from the combination of the input type, output name, and the default agent state path in a container. For example `unix:///usr/share/elastic-agent/state/data/tmp/$inputType-$outputName.sock` or `unix:///usr/share/elastic-agent/state/data/tmp/kubernetes-metrics-60f88f50-c873-11ed-9baf-09fb5640c56a.sock` for a concrete example

We have to explain a significant number of internal implementation details to document this properly, I would rather not try to document this and just prevent the agent from hitting this case at all.

The limitation here is documented in https://man7.org/linux/man-pages/man7/unix.7.html:

>   Address format
>       A UNIX domain socket address is represented in the following
>       structure:
>
>           struct sockaddr_un {
>               sa_family_t sun_family;               /* AF_UNIX */
>               char        sun_path[108];            /* Pathname */
>           };
>
>       The sun_family field always contains AF_UNIX.  On Linux, sun_path
>       is 108 bytes in size; see also NOTES, below.

The limit we use is 104 characters because the exact limit can vary based on the implementation.

I would vote we revert this, and just rely on a code fix unless we find this becomes a widespread problem. We only have one documented instance of this problem so far. @jlind23 thoughts?

jlind23: I'm torn. I would rather keep it for now until the agent behaviour is fixed just to make sure this limitation is documented. Maybe we should rather revert it and instead create a "known issue" on our end. @cmacknz thoughtS?

cmacknz: People hitting this problem are more likely to find it as a known issue as well.

cmacknz: The troubleshoot page would work, I was thinking the release notes but either works. We don't have an original issue. In the output of `elastic-agent status --output=yaml` you see something like:

```
- id: kubernetes/metrics-60f88f50-c873-11ed-9baf-09fb5640c56a
  state:
    state: 4
    message: 'Failed: pid ''3770789'' exited with code ''1'''
    units:
      ? unittype: 1
        unitid: kubernetes/metrics-60f88f50-c873-11ed-9baf-09fb5640c56a
      : state: 4
        message: 'Failed: pid ''3770789'' exited with code ''1'''
      ? unittype: 0
        unitid: kubernetes/metrics-60f88f50-c873-11ed-9baf-09fb5640c56a-kubernetes/metrics-kubelet-0d1f291d-9b2e-4f44-a0dc-82ebee865799
      : state: 4
        message: 'Failed: pid ''3770789'' exited with code ''1'''
      ? unittype: 0
        unitid: kubernetes/metrics-60f88f50-c873-11ed-9baf-09fb5640c56a-kubernetes/metrics-kube-proxy-0d1f291d-9b2e-4f44-a0dc-82ebee865799
      : state: 4
        message: 'Failed: pid ''3770789'' exited with code ''1'''
    features_idx: 0
    version_info:
      name: ""
      version: ""
```

This is accompanied by an error message in the logs:

```json
logs/elastic-agent-20230530-23.ndjson:{"log.level":"error","@timestamp":"2023-05-30T11:42:46.776Z","message":"Exiting: could not start the HTTP server for the API: listen unix /tmp/elastic-agent/6cc33ebe6ad93d3476c25a9ea38c4fe5e4c5ffbd8866f26288d86d2f672c1ea7.sock: bind: no such file or directory","component":{"binary":"metricbeat","dataset":"elastic_agent.metricbeat","id":"kubernetes/metrics-60f88f50-c873-11ed-9baf-09fb5640c56a","type":"kubernetes/metrics"},"log":{"source":"kubernetes/metrics-60f88f50-c873-11ed-9baf-09fb5640c56a"},"log.origin":{"file.line":1142,"file.name":"instance/beat.go"},"service.name":"metricbeat","ecs.version":"1.6.0","ecs.version":"1.6.0"}
```

The only work around is to reduce the length of the output name until the problem stops happening. It is related to the length of an internally generated file path as I mentioned above.

kilfoyle:
I completely forgot that our release notes have 'known issues'. Since the file length problem is planned to be resolved soon I think that's a better spot than the troubleshooting page. Here's the new PR: https://github.com/elastic/ingest-docs/pull/246

PR #257, opened 2023-06-09
Add issue template for docs requests

jrodewig:
Adds a general purpose issue template for docs requests. When someone creates an issue using the template, it presents an issue form. The form captures all information required for the issue to be ready for work by a tech writer.

The template is copied and slightly modified from the [Obs docs repo](https://github.com/elastic/observability-docs/blob/main/.github/ISSUE_TEMPLATE/other-request.yaml).

### Issue form example

Screenshot 2023-06-09 at 3 57 49 PM

### DIFF
--- a/.github/ISSUE_TEMPLATE/docs-request.yaml
+++ b/.github/ISSUE_TEMPLATE/docs-request.yaml
@@ -1,0 +1,45 @@
+name: "[Internal] Dcumentation request"
+description: Request a documentation change or enhancement for any supported release (Elastic employees).
+title: "[REQUEST]: "
+labels: ["request", "Team:Docs"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to fill out this request! This form will create an issue that the Platform Docs team will triage and prioritize.
+  - type: textarea
+    id: description
+    attributes:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/docs-request.yaml
```diff
@@ -0,0 +1,45 @@
...
+      placeholder: How can the docs team learn more about this feature?
+    validations:
+      required: true
+  - type: dropdown
+    id: collaboration
+    attributes:
+      label: Collaboration
+      description: Choose the expected collaboration model
+      options:
+        - "The documentation team will investigate the issue and create the initial content."
```
kilfoyle, line 32:
```suggestion
        - "Choose a collaboration model."
        - "The documentation team will investigate the issue and create the initial content."
```
I think we should avoid `The documentation team will investigate the issue and create the initial content` as the default. If people opening issues want writers to be responsible for the investigation and initial draft, it seems best to me that they actively, explicitly select that option.

kilfoyle (approved):
LGTM! 🚀 
Thanks a lot for adding this @jrodewig!

karenzone (approved):

PR #263, opened 2023-06-13
Add docs for per-integration health reporting

kilfoyle:
Draft only


jillguyonnet:
@kilfoyle The new section looks great, thank you! 🎉 

I would suggest the below two screenshots instead of the current ones. The latter don't have the latest UX changes and there is an inconsistency in the Overview pane (agent status should be Unhealthy, this might be a consequence of the bug I was experiencing at the time of the demo). The new ones should also be of better quality.

Screenshot 2023-06-14 at 16 28 37

Screenshot 2023-06-14 at 16 35 54

kilfoyle:
Thanks @jillguyonnet! The new screencaps are perfect. I've added them in place of the originals. The changes will show up in the [docs preview](https://ingest-docs_263.docs-preview.app.elstc.co/guide/en/fleet/master/monitor-elastic-agent.html#view-agent-details) in about 20 minutes or so, after the `elasticsearch-ci/docs` job completes.

PR #277, opened 2023-06-23
adding initcontainers explanation

gizas:
Adding some details for the initContainers as per customer request

### DIFF
--- a/docs/en/ingest-management/elastic-agent/configuration/autodiscovery/kubernetes-hints-autodiscover.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/configuration/autodiscovery/kubernetes-hints-autodiscover.asciidoc
@@ -92,7 +92,7 @@ https://github.com/elastic/elastic-agent/tree/main/deploy/kubernetes/elastic-age
 [discrete]
 == Configure hints autodiscovery
 
-To enable hints, you must add `hints.enabled: true` to the provider's configuration of https://github.com/elastic/elastic-agent/blob/main/deploy/kubernetes/elastic-agent-standalone-kubernetes.yaml[{agent}-standalone manifest] :
+To enable hints, you must add `hints.enabled: true` to the provider's configuration:
 
 [source,yaml]
 ----
--- a/docs/en/ingest-management/elastic-agent/configuration/autodiscovery/kubernetes-hints-autodiscover.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/configuration/autodiscovery/kubernetes-hints-autodiscover.asciidoc
@@ -115,7 +115,8 @@ volumes:
 ...
 ----
 
-Uncomment the init container section to download the hints templates:
+An init container is also required to download the hints templates.
+The init container is already defined, so uncomment the respective section:
 
 ["source", "yaml", subs="attributes"]
 ----
--- a/docs/en/ingest-management/elastic-agent/configuration/autodiscovery/kubernetes-hints-autodiscover.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/configuration/autodiscovery/kubernetes-hints-autodiscover.asciidoc
@@ -92,7 +92,7 @@ https://github.com/elastic/elastic-agent/tree/main/deploy/kubernetes/elastic-age
 [discrete]
 == Configure hints autodiscovery
 
-To enable hints, you must add `hints.enabled: true` to the provider's configuration:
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/elastic-agent/configuration/autodiscovery/kubernetes-hints-autodiscover.asciidoc
```diff
@@ -92,7 +92,7 @@ https://github.com/elastic/elastic-agent/tree/main/deploy/kubernetes/elastic-age
 [discrete]
 == Configure hints autodiscovery
 
-To enable hints, you must add `hints.enabled: true` to the provider's configuration:
+To enable hints, you must add `hints.enabled: true` to the provider's configuration of https://github.com/elastic/elastic-agent/blob/main/deploy/kubernetes/elastic-agent-standalone-kubernetes.yaml[{agent}-standalone manifest] :
```
ChrsMark, line 95:
```suggestion
To enable hints, you must add `hints.enabled: true` to the provider's configuration:
```
I don't see a reason pointing to a specific manifest here since the way that users can deploy Agent can be trough multiple channels (eck, helm etc).
But if we think we need to add this one you will need to fix the link to not point to the `main` branch but to the respective release's one. The reason for this is that if we update `main` for example then all the previous docs releases will be pointing to main branch. This will be inconsistent.

docs/en/ingest-management/elastic-agent/configuration/autodiscovery/kubernetes-hints-autodiscover.asciidoc
```diff
@@ -115,8 +115,7 @@ volumes:
 ...
 ----
 
-An init container is also required to download the hints templates.
-The init container is already defined, so uncomment the respective section:
+Uncomment the init container section to download the hints templates:
```
ChrsMark, line 118: Why is this change needed? It's too directive and not 100% accurate. The init container is responsible for "downloading" the templates but the way that it's phrased in this change is like "execute this command to download" which does not reflect the reality.

docs/en/ingest-management/elastic-agent/configuration/autodiscovery/kubernetes-hints-autodiscover.asciidoc
```diff
@@ -134,6 +133,10 @@ initContainers:
       mountPath: /etc/elastic-agent/inputs.d
 ----
 
+
+NOTE: The {agent} can read from `{path.config}/inputs.d`  multiple configuration files and finally produce a unified one (refer to <<elastic-agent-configuration>>). Users have the ability to manually mount their own templates under `/etc/elastic-agent/inputs.d` if they want to skip enabling initContainers section.
```
ChrsMark, line 137:
```suggestion
NOTE: The {agent} can load multiple configuration files from `{path.config}/inputs.d`  and finally produce a unified one (refer to <<elastic-agent-configuration>>). Users have the ability to manually mount their own templates under `/etc/elastic-agent/inputs.d` if they want to skip enabling initContainers section.
```

Is `<<elastic-agent-configuration>>` the link to the specific section that refers to the `inputs.d` "feature"? If we can point to this specific part it would really handy.

ChrsMark (changes requested):
Nice initiative! We need to ensure the linking within the page is correct before merging.

docs/en/ingest-management/elastic-agent/configuration/autodiscovery/kubernetes-hints-autodiscover.asciidoc
gizas: Reverted

docs/en/ingest-management/elastic-agent/configuration/autodiscovery/kubernetes-hints-autodiscover.asciidoc
gizas: This is the link of elastic-agent-configuration.asciidoc file

ChrsMark: I was mostly thinking of linking directly to docs  page part that refers the `inputs.d` usage. But I see that `inputs.d` is only mentioned in https://www.elastic.co/guide/en/fleet/current/elastic-agent-configuration.html and under no specific tittle/section so it's not possible to link directly to its section. Anyways we can add it as is and improve it further in the future is there are asks for it.

ChrsMark (approved):

PR #332, opened 2023-07-14
IngestArch: Differentiate between proxies

### LINKED ISSUE #70: IngestArch: Flesh out LS use cases

karenzone:
Closes #70

Sets up  distinction between "Logstash as proxy" and generic proxy.

### DIFF
--- a/docs/en/ingest-arch/14-ea-ls-es.asciidoc
+++ b/docs/en/ingest-arch/14-ea-ls-es.asciidoc
@@ -1,5 +1,5 @@
 [[ls-networkbridge]]
-=== {agent} to {ls} to {es}: {ls} as proxy to bridge networks
+=== {agent} to {ls} to {es}: {ls} as a proxy
 
 image::images/ea-ls-bridge.png[Image showing {agent}s collecting data and sending to {ls} for proxying before sending on to {es}]
 
--- a/docs/en/ingest-arch/14-ea-ls-es.asciidoc
+++ b/docs/en/ingest-arch/14-ea-ls-es.asciidoc
@@ -1,5 +1,5 @@
 [[ls-networkbridge]]
-=== {agent} to {ls} to {es}: {ls} to bridge networks
+=== {agent} to {ls} to {es}: {ls} as proxy to bridge networks
 
 image::images/ea-ls-bridge.png[Image showing {agent}s collecting data and sending to {ls} for proxying before sending on to {es}]
 
--- a/docs/en/ingest-arch/15-proxy.asciidoc
+++ b/docs/en/ingest-arch/15-proxy.asciidoc
@@ -6,8 +6,7 @@ image::images/ea-proxy-fs-es.png[Image showing connections between {agent} and {
 image::images/ea-fs-proxy-es.png[Image showing connections between {agent} and {es} using a proxy]
 
 TIP: This architecture works with a variety of proxying tools to allow for more flexibility in your environment.
-Consider using <<ls-networkbridge,{ls} as proxy>> when it is feasable for your use case.
-{ls} and the rest of the {stack} are designed and tested to work together, and can be easier to support. 
+Consider using <<ls-networkbridge,{ls} as a proxy>>, as {ls} and the rest of the {stack} are designed and tested to work together, and can be easier to support.  
 
 Ingest model::
  +
--- a/docs/en/ingest-arch/15-proxy.asciidoc
+++ b/docs/en/ingest-arch/15-proxy.asciidoc
@@ -5,6 +5,10 @@ image::images/ea-proxy-fs-es.png[Image showing connections between {agent} and {
 
 image::images/ea-fs-proxy-es.png[Image showing connections between {agent} and {es} using a proxy]
 
+TIP: This architecture works with a variety of proxying tools to allow for more flexibility in your environment.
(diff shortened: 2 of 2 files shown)


docs/en/ingest-arch/14-ea-ls-es.asciidoc
```diff
@@ -1,5 +1,5 @@
 [[ls-networkbridge]]
-=== {agent} to {ls} to {es}: {ls} to bridge networks
+=== {agent} to {ls} to {es}: {ls} as proxy to bridge networks
```
karenzone, line 2:
```suggestion
=== {agent} to {ls} to {es}: {ls} as proxy
```

jrodewig: Nit: Should there be an "a" here?
```suggestion
=== {agent} to {ls} to {es}: {ls} as a proxy
```

karenzone: This commit is intended to compare LS-as-proxy with the generic proxy architecture. To that end, should I shorten the title to emphasize LS as proxy for a broader range of applications?  

We could continue outlining the use case to bridge networks, and then possibly flesh out additional use cases if that makes sense.

docs/en/ingest-arch/15-proxy.asciidoc
```diff
@@ -5,6 +5,10 @@ image::images/ea-proxy-fs-es.png[Image showing connections between {agent} and {
 
 image::images/ea-fs-proxy-es.png[Image showing connections between {agent} and {es} using a proxy]
 
+TIP: This architecture works with a variety of proxying tools to allow for more flexibility in your environment.
+Consider using <<ls-networkbridge,{ls} as proxy>> when it is feasable for your use case.
```
jrodewig, line 9: Small typo
```suggestion
Consider using <<ls-networkbridge,{ls} as proxy>> when it is feasible for your use case.
```
🥜 gallery comment: Can we make this tip a bit more actionable? Why might it be feasible or not? What should the user consider?

I don't think we need to make the tip lengthy. A sentence or two like you have here is enough. I'm just not sure there's enough substance in the current tip to be helpful.

I could easily be missing something. Ignore me if so.

jrodewig (commented):
Minor comments. Feel free to ignore as wanted. Thanks!

docs/en/ingest-arch/15-proxy.asciidoc
karenzone: It's a bit tricky, and I'm hoping to get additional insights from @jsvd and @flexitrev.  
We like customers to use our products as much as possible, and it's easier for us to support customers when they build solutions using our products. 

This architecture allows for situations where users are heavily invested  in other vendors and products, and it's easier for them to go with what they know.

jrodewig: > We like customers to use our products as much as possible, and it's easier for us to support customers when they build solutions using our products.

Can we just recommend using Logstash as a proxy and note better support as a key benefit? Again — Feel free to ignore me if needed. Thanks!

docs/en/ingest-arch/15-proxy.asciidoc
```diff
@@ -5,6 +5,10 @@ image::images/ea-proxy-fs-es.png[Image showing connections between {agent} and {
...
+{ls} and the rest of the {stack} are designed and tested to work together, and can be easier to support.
```
jsvd, line 10:
```suggestion
TIP: This architecture works with a variety of proxying tools to allow for more flexibility in your environment.
Consider using <<ls-networkbridge,{ls} as a proxy>> since {ls} and the rest of the {stack} are designed and tested to work together, and can be easier to support. 
```

flexitrev (approved):

PR #359, opened 2023-07-28
Update minimal install docs to mention tested version was 8.8

kilfoyle:
For now, our testing data goes up to version 8.8.


kilfoyle:
Closing unmerged. I'll remove the version marker altogether, and make that change in main and then backport to other releases.

PR #376, opened 2023-08-02
Note that agent needs to be restarted if the upgrade happened via RPM or DEB

nimarezainia:
Note that agent needs to be restarted if the upgrade happened via RPM or DEB

### DIFF
--- a/docs/en/ingest-management/commands.asciidoc
+++ b/docs/en/ingest-management/commands.asciidoc
@@ -461,7 +461,8 @@ manager like systemd.
 
 IMPORTANT: If you installed {agent} from a DEB or RPM package, the `install`
 command will skip the installation itself and function as an alias of the
-<<elastic-agent-enroll-command,`enroll` command>> instead.
+<<elastic-agent-enroll-command,`enroll` command>> instead. Note that after
+an upgrade of the {agent} via DEB or RPM the {agent} service would need to be restarted.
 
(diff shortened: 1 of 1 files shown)


nimarezainia:
@kilfoyle I am not sure if this is being mentioned elsewhere. Couldn't really find it.

docs/en/ingest-management/commands.asciidoc
```diff
@@ -461,7 +461,8 @@ manager like systemd.
 
 IMPORTANT: If you installed {agent} from a DEB or RPM package, the `install`
 command will skip the installation itself and function as an alias of the
-<<elastic-agent-enroll-command,`enroll` command>> instead.
+<<elastic-agent-enroll-command,`enroll` command>> instead. Note that after
+an upgrade of the {agent} via DEB or RPM the {agent} service would need to be restarted.
```
kilfoyle, line 465:
```suggestion
an upgrade of the {agent} using DEB or RPM the {agent} service needs to be restarted.
```

kilfoyle (approved):
LGTM! Thanks Nima. I'll merge and backport this into the current docs.

I think this is the perfect spot to add the restart requirement. Eventually we'll need to have upgrade instructions in the docs, and when that happens (whenever I get to it) I'll make sure that the restart requirement is mentioned.

PR #394, opened 2023-08-04
IngestArch: Add port numbers for ea-ls and ea-mq architectures and refine

karenzone:
Related: 
* #197 
* #390

**PREVIEW:** https://ingest-docs_394.docs-preview.app.elstc.co/guide/en/ingest/master/use-case-arch.html

#### Tracking

- [x] Elastic Agent to Elasticsearch: Agent installed 
- [x] Elastic Agent to Elasticsearch: APIs for collection 
- [x] Elastic Agent to Logstash for enrichment
- [x] Elastic Agent to Logstash for buffering
- [x] Elastic Agent to Logstash as a proxy
- [x] Elastic Agent to Logstash to multiple destinations
- [x] Elastic Agent to proxy to Elasticsearch (Fleet Server on Cloud)
- [x] Elastic Agent to proxy to Elasticsearch (Self-managed Fleet Server)
- [x] Elastic Agent with Kafka middleware (Logstash)
- [x] Elastic Agent with Kafka middleware (ES Sink)

PR #401, opened 2023-08-09
IngestArch: Update agent-installed graphic

karenzone:
Follow-up to #394.  Image was finalized, but not included in #394.


karenzone:
@karenzone 44193

PR #412, opened 2023-08-16
Include profiling in air-gapped env

inge4pres:
We add the Universal Profiling agent download URL to the list of artifacts to download in air-gapped environments. We also amend the list with the ability to download ARM64 binaries.


kilfoyle (approved):
LGTM! Thanks for adding this @inge4pres.
I added a label to backport this into the current 8.9 docs, but feel free to remove it if that's not what you'd like.

PR #442, opened 2023-08-30
Update air-gapped.asciidoc

predogma:
This needs to cross reference the air-gapped setup for security endpoints.

This is commonly missed with elastic-agents using Elastic Defend integtrations


kilfoyle:
LGTM! 🚀 
Thanks a lot for adding this @predogma

I added a small change so the tip will render like this:
Screenshot 2023-08-30 at 10 29 19 AM

Also, I added labels so that this will automatically backport into the current Fleet & Agent docs.

kilfoyle (approved):

PR #453, opened 2023-09-05
[WIP] Add Elasticsearch output settings to Fleet managed docs

joshdover:
Some quick copy/paste - needs cleanup and fixing

### DIFF
--- a/docs/en/ingest-management/fleet/fleet-settings-output-elasticsearch.asciidoc
+++ b/docs/en/ingest-management/fleet/fleet-settings-output-elasticsearch.asciidoc
@@ -56,7 +56,7 @@ To learn about proxy configuration, refer to <<fleet-agent-proxy-support>>.
 
 | YAML settings that will be added to the {es} output section of each policy
 that uses this output. Make sure you specify valid YAML. The UI does not
-currently provide validation.
+currently provide validation. See below for which settings are available.
 
 // =============================================================================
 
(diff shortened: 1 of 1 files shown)


kilfoyle:
@joshdover  I'm happy to fix this one up if you like.

joshdover:
@kilfoyle That would be amazing. I wanted to open this before I forgot.

I think we should also do the same for Logstash and Kafka as well.

kilfoyle:
Will do! 👍  
A couple of questions: 

* This PR copies over some of the settings listed in the [Performance tuning settings](https://www.elastic.co/guide/en/fleet/current/elasticsearch-output.html#output-elasticsearch-performance-tuning-settings) section of the standalone agent Elasticsearch output page (`backoff.init`, `backoff.max`, `bulk_max_size`, `max_retries`, `timeout`). Are there any other settings that should be included?

* For Logstash, would it be the same list?

* For Kafka, we don't have any output settings in the Standalone Agent docs, like we do for Elasticsearch and Logstash. Is there somewhere I can find out what settings could go into the Advanced YAML configuration?

nimarezainia (approved):

joshdover:
> * This PR copies over some of the settings listed in the [Performance tuning settings](https://www.elastic.co/guide/en/fleet/current/elasticsearch-output.html#output-elasticsearch-performance-tuning-settings) section of the standalone agent Elasticsearch output page (`backoff.init`, `backoff.max`, `bulk_max_size`, `max_retries`, `timeout`). Are there any other settings that should be included?

These and the other ones included in the doc (eg. the reference to the common `worker` option) I think are all that are supported.

> * For Logstash, would it be the same list?

For Logstash we support all of the settings the Filebeat supports: https://www.elastic.co/guide/en/beats/filebeat/current/logstash-output.html

Only caveat is that we shouldn't mention settings that are already covered by the existing UI controls, only options that aren't available in the UI. Let me know if you need some assistance figuring out which those are.

> * For Kafka, we don't have any output settings in the Standalone Agent docs, like we do for Elasticsearch and Logstash. Is there somewhere I can find out what settings could go into the Advanced YAML configuration?

Same for Kafka as Logstash: https://www.elastic.co/guide/en/beats/filebeat/current/kafka-output.html

We also need to omit any options that are already covered by the UI

kilfoyle:
Thanks for the info!
I thought it would be a bit easier to open a new PR, so please have a look here: https://github.com/elastic/ingest-docs/pull/465

nimarezainia (approved):

nimarezainia:
> We also need to omit any options that are already covered by the UI

@joshdover do you have an example of this? I can't see any of these that are covered by the UI (yet)

joshdover:
> @joshdover do you have an example of this? I can't see any of these that are covered by the UI (yet)

In the Logstash example, `hosts`, `ssl`, `proxy_*` are already covered by the UI

nimarezainia:
> > @joshdover do you have an example of this? I can't see any of these that are covered by the UI (yet)
> 
> In the Logstash example, `hosts`, `ssl`, `proxy_*` are already covered by the UI

ok thanks. I thought that there was something there by mistake today that needed to be removed.

kilfoyle:
I think this can be closed now in favour of https://github.com/elastic/ingest-docs/pull/465 which has been merged.

PR #461, opened 2023-09-05
Update the comparison table with Kafka support

nimarezainia:


kilfoyle (approved):
LGTM!

kilfoyle:
Looks good! I updated the labels as I think this should go into the 8.10 docs, but correct me if that's wrong.

PR #479, opened 2023-09-12
Add option enrich none to input

herrBez:
Add [enrich](https://www.elastic.co/guide/en/logstash/current/plugins-inputs-elastic_agent.html#plugins-inputs-elastic_agent-ecs_metadata) option to none in the configuration example to prevent Logstash from adding the field like `event.original`, which can break integrations like apache.access (which expects the field event.original to not exist before the ingest pipeline). This should also slightly improve input performances, because it avoids Logstash to perform unuseful work.

An alternative configuration could be:
```
input {
  elastic_agent {
    port => 5044
    enrich => [ssl_peer_metadata, source_metadata]
  }
}
```

### DIFF
--- a/docs/en/ingest-management/elastic-agent/configuration/outputs/output-logstash.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/configuration/outputs/output-logstash.asciidoc
@@ -42,7 +42,9 @@ incoming {agent} connections and routes received events to {es}:
 input {
   elastic_agent {
     port => 5044
-    enrich => none
+    enrich => none # don't modify the events' schema at all
+    # or minimal change, add only ssl and source metadata
+    # enrich => [ssl_peer_metadata, source_metadata]
   }
 }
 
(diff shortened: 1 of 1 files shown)


herrBez:
Related issues: [#3451](https://github.com/elastic/integrations/issues/3451#issue-1255575421). Maybe we can close this until it is cleared the situation.

docs/en/ingest-management/elastic-agent/configuration/outputs/output-logstash.asciidoc
```diff
@@ -42,6 +42,7 @@ incoming {agent} connections and routes received events to {es}:
 input {
   elastic_agent {
     port => 5044
+    enrich => none
```
jsvd, line 45:
```suggestion
    enrich => none # don't modify the events' schema at all
    # or minimal change, add only ssl and source metadata
    # enrich => [ssl_peer_metadata, source_metadata]
```

jsvd (approved):
LGTM

herrBez:
Hi @jsvd and @karenzone can I merge the issue? :)

karenzone (approved):
LGTM. Thanks for this contribution.

karenzone:
@herrBez: Yes, merge at will. Thanks!

PR #484, opened 2023-09-13
Fix `welcome-to-elastic` links

jrodewig:
**Problem:** In https://github.com/elastic/docs/pull/2752, we updated the URL prefix (`welcome-to-elastic`) and name for the "Welcome to Elastic Docs" docs. However, we still have some stray links that use the old `/welcome-to-elastic` URL prefix

**Solution:** Updates several outdated links to use an attribute.


kilfoyle (approved):
LGTM! 🚀

PR #495, opened 2023-09-14
rename php to pgp in troubleshooting.asciidoc

essodjolo:


kilfoyle (approved):
Thanks a lot for catching and fixing that, @essodjolo!

kilfoyle:
I'll merge and make sure this fix goes into the current docs.

Issue #521, opened 2023-09-28
Add account and permissions requirements for Elastic Agent


kilfoyle:
More detailed requirements are needed in the Elastic Agent install docs, for both Linux/Unix and Windows, as requested by our SAs.

The requirements will also need to be updated once https://github.com/elastic/ingest-dev/issues/1766 is closed.

kilfoyle:
Closing. We'll find another route to assist the customers associated with this request.

PR #522, opened 2023-09-28
Add account and permissions requirements for Elastic Agent

kilfoyle:
This adds details about the account and permissions requirements to install and run Elastic Agent.

This draft is based entirely off of an email thread with the SAs so it definitely needs technical input. Any suggestions for additions and changes are very welcome!

[PREVIEW PAGE](https://ingest-docs_522.docs-preview.app.elstc.co/guide/en/fleet/master/install-agent-account-requirements.html)

Closes: #521

### DIFF
--- a/docs/en/ingest-management/elastic-agent/install-agent-account-requirements.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/install-agent-account-requirements.asciidoc
@@ -1,0 +1,60 @@
+[[install-agent-account-requirements]]
+= {agent} account and permissions requirements
+
+There are a number of account and permissions requirements that should be met before running {agent} on your system.
+
+* <<elastic-agent-account-requirements-linux>>
+* <<elastic-agent-account-requirements-windows>>
+
+[discrete]
+[[elastic-agent-account-requirements-linux]]
+== Linux and Unix requirements
+
+Currently, on Linux and Unix systems {agent} needs to run as a super user with `root` privileges. 
+
+As well, {agent} components (Beats) each use a unique account with specific controlled `root` privileges, as shown in the following table.
+
+.Linux and Unix requirements for agent components
+|===
+|Component |Function |Requirements
+
+|{auditbeat}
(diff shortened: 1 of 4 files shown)


kilfoyle:
Pasting in an update from @jamie-p-pope:

"yeah this is a good start, we're going to have to build on it though ... after creating a user and adding it to the sudoers group, the documentation needs to continue on how to actually "run the agent" as said user ... it will need to also include any directory permissions changes needed (e.g /var/log) and any updates that need to be made to systemd etc... "

docs/en/ingest-management/elastic-agent/install-agent-account-requirements.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+
+|{filebeat}
+|Forwards and centralizes your log data.
+a|* Permissions to read logs in `/var/log`.
+* Permissions to read and write logs in cache to `/var/log/Elastic/.
+
+|{metricbeat}
+|Collect metrics from operating systems and services running on your servers.
+a|* `sudo` privilege to run commands to check disk IO, network IO, memory, process, and CPU level statistics.
+* Permissions to read logs in `/var/log`.
```
cmacknz, line 60: I don't think Metricbeat reads logs in general. This line is repeated for each Beat, but I'm not sure where it came from.

nimarezainia:
@cmacknz @blakerouse , David has put this page together based on an email that came from SAs, sounds like what they are asking for is just not supported by us at this point.  If that is true can be just junk this PR and deal with the customer issue some other way? thanks

cmacknz:
It's not clear what the original request was, but if it is to allow a user to control which user the Elastic Agent runs as that is an enhancement request and not something we can fix via docs.

kilfoyle:
Thanks very much for the input on this Craig and Nima!

I know this PR was very vague. The matter discussed in email can hopefully be resolved with the individual customers, and if we need updates to the docs we can arrange that through a GitHub issue, as usual.

Closing...

PR #160, opened 2023-04-19
Add instructions on customizing installation base path for Elastic Agent

ycombinator:
This PR modifies the Installation Layout documentation for Elastic Agent.  It adds instructions to the MacOS, Linux, and Windows tabs on customizing the installation base path. 

As of this writing, we do not have a way to customize the installation base path for Debian or RPM packaged Elastic Agents.  So the instructions in those two tabs remain unchanged.

Corresponding implementation PR: https://github.com/elastic/elastic-agent/pull/2500

PR #532, opened 2023-10-04
Re-add docs for install '--base-path' parameter

kilfoyle:
This is essentially a re-do of https://github.com/elastic/ingest-docs/pull/160 which was reverted.

In https://github.com/elastic/ingest-dev/issues/1609 the install command's `--base-path` option was temporarily hidden. It's re-enabled now and should also be re-added to the documentation. See [comment](https://github.com/elastic/ingest-dev/issues/1609#issuecomment-1533505729) and [comment](https://github.com/elastic/ingest-dev/issues/1609#issuecomment-1746078531).

@ycombinator I've tagged this to go into 8.10 and later docs. Do you know if it should go into 8.9 as well?

PREVIEWS

---

Linux

linux

Mac

mac

Windows

windows

### DIFF
--- a/docs/en/ingest-management/tab-widgets/install-layout.asciidoc
+++ b/docs/en/ingest-management/tab-widgets/install-layout.asciidoc
@@ -14,6 +14,8 @@ Log files for {beats} shippersfootnote:lognumbering[]
 `/usr/bin/elastic-agent`::
 Shell wrapper installed into PATH
 
+You can install {agent} in a custom base path other than `/Library`.  When installing {agent} with the `./elastic-agent install`
+command, use the `--base-path` CLI option to specify the custom base path.
 // end::mac[]
 
 // tag::linux[]
@@ -31,6 +33,8 @@ Log files for {beats} shippers
 `/usr/bin/elastic-agent`::
 Shell wrapper installed into PATH
 
+You can install {agent} in a custom base path other than `/opt`.  When installing {agent} with the `./elastic-agent install`
+command, use the `--base-path` CLI option to specify the custom base path.
 // end::linux[]
 
 // tag::win[]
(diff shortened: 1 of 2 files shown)


docs/en/ingest-management/tab-widgets/install-layout.asciidoc
```diff
@@ -46,6 +50,8 @@ Log files for {agent}footnote:lognumbering[]
 `C:\Program Files\Elastic\Agent\data\elastic-agent-*\logs\default\*-json.log*`::
 Log files for {beats} shippers
 
+You can install {agent} in a custom base path other than `C:\Program Files`.  When installing {agent} with the `./elastic-agent install`
```
ycombinator, line 53: Fixing path separator and adding extension, since it's Windows. I had it wrong in the original PR (https://github.com/elastic/ingest-docs/pull/160) as well.  But what I have below is the correct invocation; see https://www.elastic.co/guide/en/fleet/current/install-standalone-elastic-agent.html.

```suggestion
You can install {agent} in a custom base path other than `C:\Program Files`.  When installing {agent} with the `.\elastic-agent.exe install`
```

ycombinator:
> @ycombinator I've tagged this to go into 8.10 and later docs. Do you know if it should go into 8.9 as well?

I'm not sure which versions this should go into.  That will depend on which versions https://github.com/elastic/elastic-agent/pull/3517 goes into.

ycombinator:
https://github.com/elastic/elastic-agent/pull/3517 has now been merged.  We plan to backport it to `8.11`.

kilfoyle:
@ycombinator It looks to me like this should be okay to merge now. Is that right?

ycombinator:
> @ycombinator It looks to me like this should be okay to merge now. Is that right?

Yes, please go ahead.  Thanks!

PR #583, opened 2023-10-11
Add standalone agent config file example: Apache HTTP Server

kilfoyle:
This adds configuration examples for Apache logs and metrics. The intent is to add similar examples for the Nginx and System Integrations, so let's make sure the format for this one looks good and then I'll proceed with the other two.

[DOCS PREVIEW](https://ingest-docs_583.docs-preview.app.elstc.co/guide/en/fleet/master/config-file-examples.html) (follow the "Apache HTTP Server" link)

Closes: #584

### DIFF
--- a/docs/en/ingest-management/elastic-agent/configuration/examples/config-file-example-apache.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/configuration/examples/config-file-example-apache.asciidoc
@@ -126,6 +126,6 @@ inputs: <6>
 <6> For available input settings, refer to <<elastic-agent-input-configuration>>.
 <7> For available input types, refer to <<elastic-agent-inputs-list>>.
 <8> Learn about <<data-streams>> for time series data.
-<9> A user-defined ID to identify the input stream. The ID for each input stream needs to be unique. Naming the ID by appending the associated `data_stream` dataset (for example `{user-defined-unique-id}-apache.status`) is a recommended practice, but any unique ID will work.
+<9> A user-defined ID to identify the individual input stream. The ID for each input stream needs to be unique. Naming the ID by appending the associated `data_stream` dataset (for example `{user-defined-unique-id}-apache.status`) is a recommended practice, but any unique ID will work.
 <10> A user-defined dataset. You can specify anything that makes sense to signify the source of the data.
 <11> Refer to {integrations-docs}/apache#metrics[Metrics] in the Apache HTTP Server integration documentation for the type of metrics collected and exported fields.
--- a/docs/en/ingest-management/elastic-agent/configuration/examples/config-file-example-apache.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/configuration/examples/config-file-example-apache.asciidoc
@@ -1,0 +1,120 @@
+[[config-file-example-apache]]
+= Config file example: Apache HTTP Server
+
+++++
+<titleabbrev>Apache HTTP Server</titleabbrev>
+++++
+
+Include these settings in your standalone {agent} `elastic-agent.yml` configuration file to ingest data from Apache HTTP server. 
+
+* <<config-file-example-apache-logs>>
+* <<config-file-example-apache-metrics>>
+
+[[config-file-example-apache-logs]]
+== Apache HTTP Server logs
+
+["source","yaml"]
+----
+outputs: <1>
+  default:
+    type: elasticsearch <2>
+    hosts:
+      - 'https://94d343762a2b465f976aa292df6a0fe1.us-central1.gcp.cloud.es.io:443'
+    username: '${ES_USERNAME}'
+    password: '${ES_PASSWORD}'
+agent:
+  download: <3>
+    sourceURI: 'https://artifacts.elastic.co/downloads/'
+  monitoring: <4>
+    enabled: true
+    use_output: default
+    namespace: default
+    logs: true
+    metrics: true
+inputs: <5>
+  - id: logfile-apache-fb9e632d-3708-4e76-bf4e-9758d379507e
+    name: apache-1
+    type: logfile <6>
+    use_output: default
+    data_stream:
+      namespace: default
+    streams:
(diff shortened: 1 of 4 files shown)


docs/en/ingest-management/elastic-agent/configuration/examples/config-file-example-apache.asciidoc
```diff
@@ -0,0 +1,126 @@
...
+[[config-file-example-apache-logs]]
+== Apache HTTP Server logs
+
+["source","yaml"]
+----
+outputs: <1>
+  default:
+    type: elasticsearch <2>
+    hosts:
+      - 'https://94d343762a2b465f976aa292df6a0fe1.us-central1.gcp.cloud.es.io:443'
```
nimarezainia, line 22: I believe this url here should be made generic and not be pointing to an actual URL. Somethign like "your elasticsearch output" should suffice.

docs/en/ingest-management/elastic-agent/configuration/examples/config-file-example-apache.asciidoc
```diff
@@ -0,0 +1,126 @@
...
+  download: <3>
+    sourceURI: 'https://artifacts.elastic.co/downloads/'
+  monitoring: <4>
+    enabled: true
+    use_output: default
+    namespace: default
+    logs: true
+    metrics: true
+inputs: <5>
+  - id: logfile-apache-fb9e632d-3708-4e76-bf4e-9758d379507e
```
nimarezainia, line 35: @kilfoyle i believe this id here (and others) just needs to be unique. perhaps it's explained in the bats docs. I wouldn't want the reader to think that they have to insert this exact id here.

@cmacknz is that correct?

docs/en/ingest-management/elastic-agent/configuration/examples/config-file-example-apache.asciidoc
```diff
@@ -0,0 +1,126 @@
...
+[[config-file-example-apache-metrics]]
+== Apache HTTP Server metrics
+
+["source","yaml"]
+----
+outputs: <1>
+  default:
+    type: elasticsearch <2>
+    hosts:
+      - 'https://94d343762a2b465f976aa292df6a0fe1.us-central1.gcp.cloud.es.io:443'
```
nimarezainia, line 88: same as above.

nimarezainia (approved):
Hi David, 
Approving the changes just those small modifications may be good to have. thanks again.

docs/en/ingest-management/elastic-agent/configuration/examples/config-file-example-apache.asciidoc
kilfoyle: Thanks. I've changed this and similar "ID" lines to `{user-defined-unique-id} <7>` and the description to `A user-defined unique ID to identify the input stream.`.

docs/en/ingest-management/elastic-agent/configuration/examples/config-file-example-apache.asciidoc
kilfoyle: I've changed this one to `'{elasticsearch-host-url}' <3>` and the description to `<3> The URL of the Elasticsearch cluster where output should be sent, including the port number. For example https://12345ab6789cd12345ab6789cd.us-central1.gcp.cloud.es.io:443.`

docs/en/ingest-management/elastic-agent/configuration/examples/config-file-example-apache.asciidoc
kilfoyle: Changed similarly to above.

docs/en/ingest-management/elastic-agent/configuration/examples/config-file-example-apache.asciidoc
```diff
@@ -0,0 +1,130 @@
...
+          type: logs
+        paths: <11>
+          - /var/log/apache2/access.log*
+          - /var/log/apache2/other_vhosts_access.log*
+          - /var/log/httpd/access_log*
+        tags:
+          - apache-access
+        exclude_files:
+          - .gz$
+      - id: {user-defined-unique-id} <7>
```
cmacknz, line 54: This ID needs to be different from the one above on line 42.

Add the data_stream dataset as a suffix is a good way to do this, so this could be `{user-defined-unique-id}-apache.error` here and `{user-defined-unique-id}-apache.access` above. Anything that results in distinct IDs works, it doesn't have to be this.

kilfoyle: Nice catch @cmacknz!

I've updated both the logs and the metrics examples. Here's the logs one:

screenshot

...

cap pn

nimarezainia: I wonder if we should just explain it, something like "insert a unique identifier here"

then provide the stylistic recommendation for the user to do {user-defined-unique-id}-apache.error

kilfoyle: @nimarezainia  I like it!

For the logging example I've changed the code example to:  

`id: "insert a unique identifier here" <10>`

And the description to: 
> <10> Specify a unique ID for each individual input stream. Naming the ID by appending the associated `data_stream` dataset (for example `{user-defined-unique-id}-apache.access` or `{user-defined-unique-id}-apache.error`) is a recommended practice, but any unique ID will work.

For the metrics example I've changed the code example to:

`id: "insert a unique identifier here" <9>`

And the description to:
> <9> Specify a unique ID for each individual input stream. Naming the ID by appending the associated `data_stream` dataset (for example `{user-defined-unique-id}-apache.status`) is a recommended practice, but any unique ID will work.

nimarezainia (approved):

cmacknz (approved):

Issue #232, opened 2023-05-29
[Fleet]: Unenrolled status is not available under mentioned view agent status screenshot


harshitgupta-qasource:
**Description:**
Unenrolled status is not available under mentioned view agent status screenshot

**Link:**
- https://www.elastic.co/guide/en/fleet/current/monitor-elastic-agent.html

**Screen-shot:**

![image_720](https://github.com/elastic/ingest-docs/assets/101545338/e9755c19-c0ed-4189-9e4c-e246be9a452c)

kilfoyle:
Sorry, re-opening this as I closed the wrong issue by mistake (I meant to close #/232 in the platform-docs-team repo instead). I'll fix this one asap.

harshitgupta-qasource:
Hi @kilfoyle

Thank you for the fixes.
We have revalidated the latest docs and found this issue **fixed now**.

**Link:**
- https://www.elastic.co/guide/en/fleet/current/monitor-elastic-agent.html


**Screenshot:**
![image](https://github.com/elastic/ingest-docs/assets/101545338/8a4296b8-a8b2-4a40-9ef5-341e3508d8e7)

Hence we are marking this issue as **QA:Validated**.

Thanks

PR #590, opened 2023-10-12
Add known issue for Elastic Defend & Agent CPU spike bug

kilfoyle:
A customer reported that this CPU spike issue wasn't documented in the Fleet & Elastic Agent Release Notes, so this fixes that based on this really nice [Knowledge Base article]( https://support.elastic.co/knowledge/72d18596).

@brian-mckinney and @gabriellandau  I think you two fixed this issue. Would you mind checking that my descriptions are correct, below? By the way, for any unusual issues like this that cross products, please let me or one of the security docs folks know and we'll make sure to update the ingest docs. Thanks!

Closes: #232

---

**Known issue in 8.8.0 and 8.8.1 Release Notes:**

knownissue

**Resolution in 8.8.2 Release Notes:**

resolved


gabriellandau (approved):
Thanks @kilfoyle

PR #599, opened 2023-10-17
Update the text around how much disk space is required by agent.

nimarezainia:
which is doubled during upgrades.


kilfoyle (approved):
LGTM! I added a slight wording change. I'll merge and backport into the current docs.

PR #641, opened 2023-11-01
Rename deployment-models to fix web redirect

kilfoyle:
It seems that there's a web redirect in place sending `https://www.elastic.co/guide/en/fleet/current/deployment-models.html` to `https://www.elastic.co/guide/en/fleet/current/add-a-fleet-server.html`

This is affecting the live docs so the fastest solution would seem to rename `deployment-models.html`.


jrodewig (approved):

docs/en/ingest-management/fleet/deployment-models.asciidoc
jrodewig, line 1: Should we remove this file now?

kilfoyle: Oops! I thought I did remove it. I'll fix that up in a second PR. Thanks for catching that.

kilfoyle: follow up PR: https://github.com/elastic/ingest-docs/pull/644

PR #488, opened 2023-09-13
Fix malformed link

jrodewig:
**Problem:** [Scaling Elastic Agent on Kubernetes](https://www.elastic.co/guide/en/fleet/current/scaling-on-kubernetes.html) contains a malformed link. Screenshot:

Screenshot 2023-09-13 at 2 46 18 PM

**Solution:** Fix the link syntax.

PR #649, opened 2023-11-03
Fixed broken link to KSM sharding

alstolten:
Hi team, it seems the link moved, fixed it.

### DIFF
--- a/docs/en/ingest-management/elastic-agent/scaling-on-kubernetes.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/scaling-on-kubernetes.asciidoc
@@ -111,7 +111,7 @@ The following **alternative configuration methods** have been verified:
   - For non-leader {agent} deployments that collect per KSM shards
 2. With `taint/tolerations` to isolate the {agent} daemonset pods from rest of deployments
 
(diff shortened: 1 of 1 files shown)


kilfoyle (approved):
LGTM! Thanks for fixing this @alstolten

Just one suggestion: We normally push changes to `main` and backport into the published release branches. After this one merges I will forward-port into the later branches so that the change doesn't get lost.

alstolten:
Thank you @kilfoyle! I just hit the edit button from the docs page this time, and the rest I did what GH suggested to me. Anyway, what your wrote makes total sense, will check next time to PR to main.

kilfoyle:
Just for reference, it looks like James already fixed this in 8.10 and main. :-) 
https://github.com/elastic/ingest-docs/pull/488

alstolten:
Hmm, #488 does not seem to change anything with regards to `https://github.com/elastic/elastic-agent/blob/main/docs/elastic-agent-ksm-sharding.md` but I might miss sth.

alstolten:
Perfect. Then we would just need to fix it for older versions.

kilfoyle:
> Perfect. Then we would just need to fix it for older versions.

I've added backport labels for 8.8 and 8.7 so that should do the trick.

Issue #653, opened 2023-11-06
Change standalone config examples to use API keys


kilfoyle:
The [config file examples](https://www.elastic.co/guide/en/fleet/current/config-file-examples.html) for standalone Elastic Agent should use API keys rather than basic auth.

> Nice to see the additional docs for standalone Elastic Agent. Can we move these docs over to use API key instead of username/password? API key should be best practice and I think is the only way to get this to work in Serverless.

PR #655, opened 2023-11-06
Change standalone agent config examples to use API key

kilfoyle:
API key authentication is required for Serverless, so we should use it as the default in our configuration examples.

Closes: #653

---

Screenshot 2023-11-06 at 6 37 03 PM
 . . . 
Screenshot 2023-11-06 at 6 08 33 PM

---

Screenshot 2023-11-06 at 6 34 44 PM

### DIFF
--- a/docs/en/ingest-management/elastic-agent/configuration/examples/config-file-example-apache.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/configuration/examples/config-file-example-apache.asciidoc
@@ -20,30 +20,29 @@ outputs: <1>
     type: elasticsearch <2>
     hosts:
       - '{elasticsearch-host-url}' <3>
-    username: '${ES_USERNAME}'
-    password: '${ES_PASSWORD}'
+    api_key: "my_api_key" <4>
 agent:
-  download: <4>
+  download: <5>
     sourceURI: 'https://artifacts.elastic.co/downloads/'
-  monitoring: <5>
+  monitoring: <6>
     enabled: true
     use_output: default
     namespace: default
     logs: true
     metrics: true
-inputs: <6>
-  - id: "insert a unique identifier here" <7>
+inputs: <7>
+  - id: "insert a unique identifier here" <8>
     name: apache-1
-    type: logfile <8>
+    type: logfile <9>
     use_output: default
-    data_stream: <9>
+    data_stream: <10>
       namespace: default
     streams:
-      - id: "insert a unique identifier here" <10>
+      - id: "insert a unique identifier here" <11>
         data_stream:
-          dataset: apache.access <11>
+          dataset: apache.access <12>
           type: logs
-        paths: <12>
+        paths: <13>
           - /var/log/apache2/access.log*
           - /var/log/apache2/other_vhosts_access.log*
           - /var/log/httpd/access_log*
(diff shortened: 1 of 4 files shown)


pierrehilbert (approved):

ruflin (commented):
Thanks for updating this for API Key. I think there are potentially some more places out there with Username / Password (k8s).

One questions users will likely have is, how can I create the API key / what permissions do I need to give it? Do we have good docs on this?

kilfoyle:
Thanks for the feedback @ruflin. I'm going to merge this PR, which is specifically for the config files, but I've opened a separate issue to deal with other instances in the Fleet and Agent docs where we use basic auth.

> One questions users will likely have is, how can I create the API key / what permissions do I need to give it? Do we have good docs on this?

The description "An [API key](https://www.elastic.co/guide/en/kibana/current/api-keys.html) used to authenticate with the Elasticsearch cluster." links to the "API keys" page in the Kibana docs, which explains how to make the key (with screencaps), security privileges, etc.

> I think there are potentially some more places out there with Username / Password (k8s).

I count 89 references to "password" in the Fleet & Agent docs. I've opened a [separate issue](https://github.com/elastic/ingest-docs/issues/693) to comb through those and, where appropriate, change them to use API key authentication.

jrodewig (approved):

ruflin:
> The description "An [API key](https://www.elastic.co/guide/en/kibana/current/api-keys.html) used to authenticate with the Elasticsearch cluster." links to the "API keys" page in the Kibana docs, which explains how to make the key (with screencaps), security privileges, etc. So I think that'll be clear for users.

@kilfoyle I'm challenging you to select the "correct" minimal permissions for an Elastic Agent, can you do it without looking it up? It is not about creating the API key, it is about giving it the correct permissions.

nimarezainia:
@kilfoyle thanks for making those changes, we do need to remove the references to username/password everywhere - so let's pursue the issue you created separately.

Here what @ruflin rightly is asking for (imo) is an extension to the idea of helping users to create their own configuration. We currently don't have any instruction for the user on how an API key should be created with the **restricted** permissions. I think we can enhance our current set of configuration examples with API creation (tutorial? for lack of a better word). For the most part the pattern is there on how to create these, similar to how the anatomy of the standalone configuration was broken down. Docs on [API creation](https://www.elastic.co/guide/en/elasticsearch/reference/8.10/security-api-create-api-key.html) do explain it pretty well, but still I reckon it's difficult to create an API key with limited permission.

As an example this is the relevant section for the nginx logs that the user would need to create:

    "indices": [
      {
        "names": [
          "logs-nginx.access-default"
        ],
        "privileges": [
          "auto_configure",
          "create_doc"
        ],
        "allow_restricted_indices": false
      },
      {
        "names": [
          "logs-nginx.error-default"
        ],
        "privileges": [
          "auto_configure",
          "create_doc"
        ],
        "allow_restricted_indices": false
      },


Let's add this to the list if you don't mind and I can get the content to you. We can figure where is the best location for it in the docs.

ruflin:
To simplify things for standalone, I would recommend users to just create API keys for `logs-*-*` and not specific to the dataset. Otherwise each time they change a config on the edge, the API key might fail. To go broader, we could add `metrics-*-*` to it too.

kilfoyle:
@ruflin, @nimarezainia Thanks for the input. I didn't understand that the API permissions are part of the standalone agent configuration, but of course that makes sense.

I've opened https://github.com/elastic/ingest-docs/issues/697 and set it to a high priority, so we can get these config examples fixed up soon.

kilfoyle:
Aha! I've just discovered this [Grant standalone Elastic Agents access to Elasticsearch](https://www.elastic.co/guide/en/fleet/current/grant-access-to-elasticsearch.html) page, so we definitely have something to work with here. Sorry for not having noticed this before.

Issue #692, opened 2023-11-16
[Fleet] Document output secrets


jillguyonnet:
This is a followup to https://github.com/elastic/ingest-docs/issues/161 which documented how Fleet handles secret in agent policies.

The scope of [secret storage phase 1](https://github.com/elastic/kibana/issues/154715) also includes storing secrets that exist in outputs: https://github.com/elastic/kibana/issues/157458.

### Details

At this time, output secrets are one of the following:
- Kafka output password
- Kafka output SSL key
- Logstash output SSL key
- Remote ES service token

There are two ways secret values can be set for outputs:
- UI (see https://github.com/elastic/kibana/pull/169429 for screenshots): on creating an output, the user can use a secret (default) or opt to use plain text instead.
- Preconfigured outputs from the Kibana config file (https://github.com/elastic/kibana/issues/166360): a secure hash of the secret value obtained with the Argon2id algorithm is stored in the output's saved object in order for Kibana to manage updates of the secret. At no point is the value itself stored in Kibana.

jillguyonnet:
Adding [remote ES outputs](https://github.com/elastic/kibana/pull/171875) to this.
cc @juliaElastic

kilfoyle:
Completed via [PR](https://github.com/elastic/kibana/pull/172637), [PR](https://github.com/elastic/ingest-docs/pull/696), [PR](https://github.com/elastic/ingest-docs/pull/762)

Thanks a lot for all the help, Jill!

PR #696, opened 2023-11-16
Docs for secrets storage in agent outputs

kilfoyle:
This adds docs for agent output secrets. 

@jillguyonnet This is definitely a "first draft". I'm not sure that I've got the "Policy secret values" section (down at the bottom) quite right. Please let me know what you think.

Target: 8.12
Rel: #692

Docs preview links:
 - [Kafka authentication settings](https://ingest-docs_696.docs-preview.app.elstc.co/guide/en/fleet/master/kafka-output-settings.html#_authentication_settings)
 - [Logstash output settings](https://ingest-docs_696.docs-preview.app.elstc.co/guide/en/fleet/master/ls-output-settings.html#ls-output-settings)
---

Update to Logstash and Kafka outputs pages:
Screenshot 2023-12-05 at 4 43 45 PM

### DIFF
--- a/docs/en/ingest-management/agent-policies.asciidoc
+++ b/docs/en/ingest-management/agent-policies.asciidoc
@@ -282,7 +282,7 @@ will still forward data to the cluster.
 
 When you create an integration policy you often need to provide sensitive information such as an API key or a password. To help ensure that data can't be accessed inappropriately, any secret values used in an integration policy are stored separately from other policy details.
 
-When you add a secret value directly to the {kibana-ref}/fleet-settings-kb.html[`kibana.yml` configuration file], the value is replaced with a reference, with the original value stored outside of {fleet} and {kib} as a secure hash. Note that this type of secret storage requires all configured {fleet-server}s to be on version 8.12.0 or higher.
+As well, after you've saved a secret value in {fleet}, the value is hidden in both the {fleet} UI and in the agent policy definition. When you view the agent policy (**Actions -> View policy**), an environment variable is displayed in place of any secret values, for example `${SECRET_0}`.
 
 When you save a secret value in {fleet}, the value is hidden in both the {fleet} UI and in the agent policy definition. When you view the agent policy (**Actions -> View policy**), an environment variable is displayed in place of any secret values, for example `${SECRET_0}`. In order for sensitive values to be stored secretly in {fleet}, all configured {fleet-server}s must be on version 8.10.0 or higher.
 
--- a/docs/en/ingest-management/agent-policies.asciidoc
+++ b/docs/en/ingest-management/agent-policies.asciidoc
@@ -284,7 +284,7 @@ When you create an integration policy you often need to provide sensitive inform
 
 As well, after you've saved a secret value in {fleet}, the value is hidden in both the {fleet} UI and in the agent policy definition. When you view the agent policy (**Actions -> View policy**), an environment variable is displayed in place of any secret values, for example `${SECRET_0}`.
 
-When you save a secret value in {fleet}, the value is hidden in both the {fleet} UI and in the agent policy definition. When you view the agent policy (**Actions -> View policy**), an environment variable is displayed in place of any secret values, for example `${SECRET_0}`. In order for sensitive values to be stored secretly in {fleet}, all configured {fleet-server}s must be on version 8.10.0 or higher.
+WARNING: In order for sensitive values to be stored secretly in {fleet}, all configured {fleet-server}s must be on version 8.10.0 or higher.
 
 Though secret values stored in {fleet} are hidden, they can be updated. To update a secret value in an integration policy:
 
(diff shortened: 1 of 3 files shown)


docs/en/ingest-management/agent-policies.asciidoc
```diff
@@ -282,9 +282,9 @@ will still forward data to the cluster.
 
 When you create an integration policy you often need to provide sensitive information such as an API key or a password. To help ensure that data can't be accessed inappropriately, any secret values used in an integration policy are stored separately from other policy details.
 
-As well, after you've saved a secret value in {fleet}, the value is hidden in both the {fleet} UI and in the agent policy definition. When you view the agent policy (**Actions -> View policy**), an environment variable is displayed in place of any secret values, for example `${SECRET_0}`.
+When you add a secret value directly to the {kibana-ref}/fleet-settings-kb.html[`kibana.yml` configuration file], the value is replaced with a reference, with the original value stored outside of {fleet} and {kib} as a secure hash. Note that this type of secret storage requires all configured {fleet-server}s to be on version 8.12.0 or higher.
```
jillguyonnet, line 285: I'm wondering whether this is the right location to add information about the `kibana.yml` config file. Right now, secret values in the `kibana.yml` file can only pertain to preconfigured outputs, all other secret values (in output or integration policies) are inputted via the UI. Since this section is under "Fleet UI settings", it might be a little confusing.

It doesn't look like there is an dedicated page for preconfigured outputs under "Manage Elastic Agents in Fleet", there are documented [here](https://www.elastic.co/guide/en/kibana/current/fleet-settings-kb.html). I'm thinking it would be good to add documentation to that page. The idea is that now a preconfigured output can specify secret values, for example in the case of a Logstash output, instead of:
```yml
xpack.fleet.outputs:
  - id: my-logstash-output
    name: preconfigured logstash output
    type:  logstash
    hosts: ["localhost:9999"]
    ssl:
      certificate: xxxxxxxxxx
      key: unsecurekey
```
We can have:
```yml
xpack.fleet.outputs:
  - id: my-logstash-output-with-a-secret
    name: preconfigured logstash output with a secret
    type:  logstash
    hosts: ["localhost:9999"]
    ssl:
      certificate: xxxxxxxxxx
    secrets:
      ssl:
        key: securekey
```

What do you think?

docs/en/ingest-management/fleet/fleet-settings-output-kafka.asciidoc
```diff
@@ -78,6 +78,8 @@ When SCRAM is enabled, {agent} uses the link:https://en.wikipedia.org/wiki/Salte
 * SCRAM-SHA-256 - uses the SHA-256 hashing function
 * SCRAM-SHA-512 - uses the SHA-512 hashing function
 
+To prevent unauthorized access your Kafka password is stored as a <<agent-policy-secret-values,secret value>>, separate from other policy details. While secret storage is recommended, you can choose to override this setting and store the password as plain text in the agent policy definition. Secret storage requires {fleet-server} version 8.12 or higher.
```
jillguyonnet, line 81: Could the link to "Policy secret value" be confusing? While the latter also store values as secrets, it's not strictly speaking related information. For the same reason, I would suggest removing ", separate from other policy details". (Same comment for Kafka and Logstash SSL keys 🙂).

docs/en/ingest-management/agent-policies.asciidoc
kilfoyle: Ah! I see. I misunderstood things. I agree that the [Fleet Settings page](https://www.elastic.co/guide/en/kibana/current/fleet-settings-kb.html) in the Kibana docs is a better home for this. I've opened https://github.com/elastic/kibana/pull/172637

Comparing that page with your example above, I noticed that the SSL settings aren't documented, so I've added them in. I also noticed that a couple of other settings are listed as required, but since they're not in your example I'm thinking they should be documented as optional. So I included that fix in the PR as well.

docs/en/ingest-management/fleet/fleet-settings-output-kafka.asciidoc
kilfoyle: Definitely. I've removed the link and the "separate from other policy details" text. The change is in both the Kafka and the Logstash sections.

kilfoyle:
Thanks for the feedback @jillguyonnet. Sorry it took me so long to circle back to this PR!

jillguyonnet (dismissed):
Changes look good, thanks!

I have a small suggestion (feel free to ignore): what do you think of adding a sentence like "Note that this setting can also be stored as secret value or plain text for preconfigured outputs" with a link to the Fleet settings page?

Also, will the secret storage of the service token for the remote Elasticsearch output type be documented separately? I had added it to the issue, but perhaps it would make more sense to include the part about secret storage in the dedicated doc issue.

karenzone (approved):
LGTM

kilfoyle:
@jillguyonnet  Great suggestions! I added the note that you suggested to the three settings mentioned in this PR.

I missed the `Remote ES service token` setting so thanks for catching that. I've opened a separate PR for that update: [Document that remote ES service token can be stored as a secret](https://github.com/elastic/ingest-docs/pull/762)

jillguyonnet (approved):
Thank you for iterating, LGTM!

karenzone (approved):
LGTM

Issue #708, opened 2023-11-21
Improve docs for Fleet support in multiple Kibana spaces 


kilfoyle:
Notes from Kyle in [sdh-beats issue](https://github.com/elastic/sdh-beats/issues/4083):

Fleet does not have first-class support for multiple spaces in general, and there will likely be issues with the majority of integrations when trying to run across multiple spaces. We have an internal meta issue here tracking the overall effort and scoping process of improving Fleet's multi-space support: https://github.com/elastic/ingest-dev/issues/1664

We mention the limitation around multiple spaces for integration assets in https://www.elastic.co/guide/en/fleet/8.11/install-uninstall-integration-assets.html#install-integration-assets.

This could probably be more explicit, and some better guidance around Fleet and Spaces would go a long way in our docs.

jalogisch:
IMHO, it should be loud and clear on the Integrations front page: https://www.elastic.co/guide/en/fleet/8.11/integrations.html

That integration might not work as expected with multiple spaces, and it most likely does not work to separate integrations into different spaces or have different configurations per space. It highly depends on the integration of what is possible and what is not.

kilfoyle:
Just for tracking, @jalogisch's PR above adds the following to the [Manage Elastic Integrations](https://www.elastic.co/guide/en/fleet/current/integrations.html) front page.

![Screenshot 2023-11-22 at 12 18 10 PM](https://github.com/elastic/ingest-docs/assets/41695641/5faf2fa0-8ffb-47af-b0f4-1d38703e6e73)

kilfoyle:
I've also updated the [Install and uninstall Elastic integration assets](https://www.elastic.co/guide/en/fleet/8.11/install-uninstall-integration-assets.html#install-integration-assets) page so the warning matches what we put on the Manage Elastic integrations front page (see comment above).

![Screenshot 2023-11-28 at 12 44 28 PM](https://github.com/elastic/ingest-docs/assets/41695641/afbd0df9-45a1-433c-867a-2d55c4a266c9)

Based on discussion with Nima it seems this is about as much as we can do for the docs right now, until we have better support for integrations with spaces, which is currently in plan.

Closing.

PR #709, opened 2023-11-22
Update integrations.asciidoc

jalogisch:
First, try to improve the docs as mentioned in https://github.com/elastic/ingest-docs/issues/708. 

This might already be enough, but we might want to improve other places.


kilfoyle (approved):
LGTM! Thanks a lot for adding this @jalogisch.
I'll merge and then forward port this into `main`.

kilfoyle:
Add: I will look into what we can add in other places in the docs.

PR #711, opened 2023-11-22
Update 8.11.1 Release Notes for upgrade issue

kilfoyle:
This updates the [8.11.1 release notes](https://www.elastic.co/guide/en/fleet/current/release-notes-8.11.1.html) to correct the known upgrade issue (the issue affects both version 8.11.0 and 8.11.1).

Screenshot 2023-11-22 at 10 25 01 AM

### DIFF
--- a/docs/en/ingest-management/release-notes/release-notes-8.11.asciidoc
+++ b/docs/en/ingest-management/release-notes/release-notes-8.11.asciidoc
@@ -41,16 +41,16 @@ Review important information about {fleet-server} and {agent} for the 8.11.1 rel
 === Known issues
 
 [[known-issue-169825-8.11.1]]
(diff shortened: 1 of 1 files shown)


kpollich:
The actual behavior here is a little more intricate. Agent version 8.11.0 is _not_ shown when on stack version 8.11.0, and agent version 8.11.0 _is_ shown when on stack version 8.11.1 😅 

Perhaps we could update the language here to capture those possibilities, or to be more generic?

kilfoyle:
@kpollich  Ah, I see. How about this? As you suggest, I think more generic phrasing is probably the way to go:


Screenshot 2023-11-22 at 11 41 11 AM

kpollich (approved):
LGTM 🚀

lcawl (approved):

PR #730, opened 2023-11-30
Add pre and post 8.11 versions of reset cloud policy command

kilfoyle:
This fixes a command on the [Troubleshooting](https://www.elastic.co/guide/en/fleet/current/fleet-troubleshooting.html#agents-in-cloud-stuck-at-updating) page. The command now has an additional header required for v8.11 and higher, so I've included the pre- and post-8.11 versions on the page.

Closes: #727

---

PR #738, opened 2023-12-05
Fix below/above terms for troubleshooting page

kilfoyle:
I merged without noticing the comments in https://github.com/elastic/ingest-docs/pull/730/files so this fixes that.


karenzone (approved):
LGTM

Issue #720, opened 2023-11-27
[REQUEST][Fleet] Document presets for performance tuning to ES output configuration


jlind23:
### Description

ES output presets will be added as part of Beats code base and we should have a doc page that redirect users these values.

### Resources

Related Kibana issue - https://github.com/elastic/kibana/issues/166870

### Collaboration

The documentation team will investigate the issue and create the initial content.

### Point of contact.

**Main contact:** @juliaElastic 

**Stakeholders:**

jlind23:
@kilfoyle adding this one to your radar

kilfoyle:
@kpollich  Before I document the (very nice) new performance presets feature I'd like to make sure that our documented [list of available advanced YAML config settings](https://www.elastic.co/guide/en/fleet/current/es-output-settings.html#es-output-settings-yaml-config) for the Elasticsearch output is up to date. I noticed that the `queue.mem.events`, `flush.min_events`, and `flush.timeout` settings aren't documented, but they're included in the presets table shown in the performance tuning issue.

Is there somewhere that I can a list of the advanced YAML configs that are available for the Elasticsearch output in Fleet?

**Comparison**

| Documented in [ES output Advanced YAML settings](https://www.elastic.co/guide/en/fleet/current/es-output-settings.html)    | Preset settings listed in [issue](https://github.com/elastic/kibana/issues/166870) |
| -------- | ------- |
| `backoff.init`  | - |
| `backoff.max` | - |
| `bulk_max_size` | `bulk_max_size` |
| `compression_level`| `compression` |
| `max_retries` | - |
| `timeout` | `idle_timeout` |
| `worker` | `workers` |
| - | `queue.mem.events` |
| - | `flush.min_events` |
| - | `flush.timeout` |

kpollich:
I believe these are the "internal queue" settings from beats, e.g. https://www.elastic.co/guide/en/beats/filebeat/current/configuring-internal-queue.html. 

@cmacknz @faec - Is there a better source of truth than the doc above that we could provide David here?

cmacknz:
Yes `queue.mem.events`, `flush.min_events`, and `flush.timeout` are Beats memory queue settings and https://www.elastic.co/guide/en/beats/filebeat/current/configuring-internal-queue.html is the best documentation for them.

PR #764, opened 2023-12-14
Updating some of the data around the tuning presets

kilfoyle:
This updates the tuning presets table to give a better indication of performance expectations.

Rel: #720


karenzone (approved):
LGTM

Issue #530, opened 2023-10-03
Document sending monitoring data to a remote Elasticsearch cluster


kilfoyle:
We'll need docs to accompany https://github.com/elastic/kibana/issues/104986, which enables sending agent monitoring to a remote Elasticsearch cluster. Feature is planned to land in 8.12.

Target is v8.12

Related docs:
 - https://www.elastic.co/guide/en/elasticsearch/reference/current/service-tokens-command.html
 - https://www.elastic.co/guide/en/elasticsearch/reference/current/service-accounts.html#service-accounts-tokens

Notes from chat with Nima:
 - Add blurb at top of Elasticsearch output settings pointing to new "Send agent monitoring data to remote ES…" page, with configuration steps.
 - Leave the "make this output the default for integrations" option as is (it won't appear in the UI anyway)
 - Remove the note about "sending to remote ES is not supported"!

PR #765, opened 2023-12-14
Remove external ES monitoring restriction

kilfoyle:
Removes "Sending monitoring data to a remote Elasticsearch cluster is currently not supported." The feature is supported as of 8.12.0.

Rel: #530


karenzone (approved):
LGTM

PR #776, opened 2023-12-19
Updating the agent vs beats compatibility

nimarezainia:
Updating the agent vs beats compatibility


karenzone (approved):
Builds cleanly and renders as expected. LGTM

PR #789, opened 2024-01-04
Add information about elastic agent diagnostics flow

philippkahr:
I think we should mention how the data is flowing. Otherwise it can be confusing. How will be the elastic agent diagnostics be retrieved, if no Elasticsearch output is defined because logstash is used.

Maybe we can add a simplified version of the flow chart: https://user-images.githubusercontent.com/6766512/190254026-75cc900f-08c6-48c7-8789-56bd8304d933.png  from https://github.com/elastic/ingest-dev/issues/1224

thanks to @ckauf for digging the things up.


kilfoyle (approved):
LGTM! Thanks for fixing this @philippkahr 
I added just one suggestion for the text.

kilfoyle (approved):
LGTM!

kilfoyle:
I'll merge this and make sure the backports go through.

PR #806, opened 2024-01-12
Update output-elasticsearch.asciidoc

nimarezainia:
the bulk-max-size and compression defaults are incorrect post 8.12

@kilfoyle I hope this is the right location to make the change that would get reflected for beats and agent. thank you.


kilfoyle (approved):

kilfoyle:
Thanks a lot @nimarezainia! I checked the Beats docs and in the `master` branch they already have the correct settings. And the settings you've updated will fix the settings in the Fleet & Agent docs.

PR #822, opened 2024-01-17
Add details about agent namespaces cascading to integration policies

kilfoyle:
With https://github.com/elastic/kibana/pull/174776, agent policy namespaces will now cascade to integration policies unless otherwise specified. This adds a paragraph to clarify that behaviour in the docs:

screen


criamico (approved):
Looks good! Will it be visible only in 8.13 docs right?

kilfoyle:
> Looks good! Will it be visible only in 8.13 docs right?

Thanks @criamico! Yes, I'll merge this into `main` with no backports, so it'll appear only in the 8.13 and later docs.

PR #871, opened 2024-01-30
Update aws-s3 known issue in 8.12 to include units in workaround

cmacknz:
Use `queue.mem.flush.timeout: 1s` instead of `queue.mem.flush.timeout: 1` to ensure the correct behavior and match the rest of our documentation.


kilfoyle (approved):
LGTM!

Issue #873, opened 2024-01-30
Update proxy docs with communication details for epr.elastic.co


kilfoyle:
A customer requested details on the communication up/down when a customer connects to [epr.elastic.co](http://epr.elastic.co/) when setting up a [proxy](https://www.elastic.co/guide/en/fleet/current/fleet-agent-proxy-support.html).

We should document the following:

> Kibana, through the Fleet plugin, is the only service interacting with the package registry in production deployments.
Communication happens when interacting with the Integrations UI, and when upgrading Kibana. The shared information is about discovery of Elastic Packages and their available versions. In general the only deployment-specific data that is shared is the kibana version.
There are some advanced (and undocumented) options that can make kibana share limited information with epr about the installed plugins, but afaik we only use this in serverless. There is no application of this for on-premise deployments.
Regarding operational data, on the package registry we collect APM data for monitoring purpouses, what can include information about the requests done. Also the public load balancers collect access logs.

PR #874, opened 2024-01-30
Add details about info shared, via proxy, with the EPR

kilfoyle:
This adds the following to the [Set the proxy URL of the Elastic Package Registry](https://www.elastic.co/guide/en/fleet/current/epr-proxy-setting.html) page.

Closes: https://github.com/elastic/ingest-docs/issues/873

---

Screenshot 2024-01-30 at 4 43 30 PM


karenzone (approved):
LGTM

PR #908, opened 2024-02-09
Add known issue for stuck agent upgrade (8.12.0, 8.12.1)

kilfoyle:
This adds a known issue for the stuck agent upgrades to the 8.12.0 and 8.12.1 Release Notes.

Thanks @julia for the really clear explanation!

Closes: #907

---

screen1
screen2

PR #918, opened 2024-02-14
Fix known issue for stuck upgrade in 8.12.1 release notes

kilfoyle:
in #908 I fixed the "stuck upgrade" known issue for 8.12.0 only, and forgot about 8.12.1. So this fixes that omission.

Also, I've added `After running these API requests, wait at least 10 minutes, and then the agents should be upgradeable again.` at the end of the workaround steps.

---

Screenshot 2024-02-14 at 11 05 50 AM
Screenshot 2024-02-14 at 11 06 01 AM


juliaElastic (approved):
Looks good, thanks!

karenzone (approved):
LGTM

Issue #693, opened 2023-11-16
Change Username/Password references to API key


kilfoyle:
API key is the preferred authentication method for Fleet & Agent to talk to Elasticsearch, so our config and other examples should be updated accordingly.

PR #924, opened 2024-02-15
Fix examples to use API key authentication

kilfoyle:
This makes a few small touch-ups to indicate that API key authentication is required on serverless, and generally recommended over basic auth. For the changes to the "debugging" example, I tested an agent policy and the resulting output from the `elastic-agent inspect` command to make sure that the updated formatting is correct.

I've also updated the API key generation to note that the UI path is slightly different in a serverless project.

The examples in the nginx monitoring tutorials were fixed separately.

Closes: #693


karenzone (approved):
LGTM

PR #929, opened 2024-02-15
Document that disk queues cannot be configured for Agent-managed Beats

ycombinator:
This PR documents explicitly that users cannot configure internal _disk_ queues for Agent-managed Beats in [this table](https://www.elastic.co/guide/en/fleet/current/beats-agent-comparison.html#supported-configurations) comparing configurations supported by Beats vs. Agent.


kilfoyle (approved):
LGTM! 🚀

PR #935, opened 2024-02-19
Update Elastic-Agent Kafka settings

belimawr:
This commit updates the documentation from the Kafka output on standalone Elastic-Agent settings by removing the options that are not supported any more.

### DIFF
--- a/docs/en/ingest-management/elastic-agent/configuration/outputs/output-kafka.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/configuration/outputs/output-kafka.asciidoc
@@ -35,8 +35,7 @@ outputs:
     partition:
       round_robin:
         group_events: 1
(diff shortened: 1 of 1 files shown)


kilfoyle (approved):
LGTM! Thanks for fixing these up @belimawr!

One question: We also have this [Processors](https://www.elastic.co/guide/en/fleet/current/kafka-output-settings.html#_topics_settings) section in the docs for Fleet-managed agent. Should that section also be removed, or are the `topics` settings only being deprecated for standalone agent?  I can look after removing the Fleet-managed section if we need that.

kilfoyle:
I just noticed [this PR](https://github.com/elastic/ingest-docs/pull/935), so I guess the settings are being removed for Fleet as well. I will update the Fleet-managed docs.

PR #973, opened 2024-03-12
Scrub book-level attributes from Fleet & Agent guide

kilfoyle:
This removes most of the book-level attributes in the Fleet & Agent guide. I'm leaving a few that have to do with the tab widgets, so for those I'll wait and see how the Observability folks fix things up.

Backporting this to 8.13 so that if I've broken anything we'll know sooner rather than later.

Closes: #974


kilfoyle:
@elasticmachine merge upstream

kilfoyle:

karenzone (approved):
LGTM

PR #981, opened 2024-03-25
[Fleet] Add variable to the preconfiguration example

nchaulet:
## Description 

Resolve https://github.com/elastic/kibana/issues/159444

Add variable to the preconfiguration example

### DIFF
--- a/docs/en/ingest-management/create-agent-policies-no-UI.asciidoc
+++ b/docs/en/ingest-management/create-agent-policies-no-UI.asciidoc
@@ -49,7 +49,7 @@ xpack.fleet.agentPolicies:
           name: fleet_server
 ----
           
-The following example creates an agent policy for general use:
+The following example creates an agent policy for general use, that customize the period for the system.core metrics:
 
 [source,yaml]
 ----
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/create-agent-policies-no-UI.asciidoc
```diff
@@ -49,7 +49,7 @@ xpack.fleet.agentPolicies:
           name: fleet_server
 ----
           
-The following example creates an agent policy for general use:
+The following example creates an agent policy for general use, that customize the period for the system.core metrics:
```
nchaulet, line 52: @kilfoyle will it make sense to add something to tell the user they cab find all the available input and variable in the integration UI?

kilfoyle: @nchaulet  I think that makes sense. Do you think it might be better to refer people to the Integrations documentation rather than the UI? Usually the docs are better as a reference, but I don't see that `period` setting described on the [System integration page](https://docs.elastic.co/en/integrations/system#).

Depending on which is better (refer people to the docs or to the UI), here are a couple of suggestions for the text.

Refer users to the docs:
```
The following example creates an agent policy for general use, and customizes the `period` setting for the `system.core` data stream. All available inputs and variables are described in the link:https://docs.elastic.co/en/integrations[Integrations] documentation.
```

Refer users to the UI:
```
The following example creates an agent policy for general use, and customizes the `period` setting for the `system.core` data stream. You can find all available inputs and variables in the **Integrations** app in {kib}.
```

nchaulet: > Do you think it might be better to refer people to the Integrations documentation rather than the UI? Usually the docs are better as a reference, but I don't see that period setting described on the [System integration page](https://docs.elastic.co/en/integrations/system#). 

Unfortunately I do not think we have the information in the Integrations documentation

kilfoyle (approved):
LGTM! 🚢 
Thanks @nchaulet!

PR #983, opened 2024-03-26
Update 8.13.0 queue documentation and release notes

cmacknz:
Bring over two relevant output changes from Beats.

Also sync the queue.flush.min_events documentation from Beats so that it is up to date with the most recent changes.

### DIFF
--- a/docs/en/ingest-management/elastic-agent/configuration/outputs/output-logstash.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/configuration/outputs/output-logstash.asciidoc
@@ -169,23 +169,34 @@ For more information, refer to <<secure-logstash-connections>>.
 
 The memory queue keeps all events in memory.
 
-The memory queue waits for the output to acknowledge or drop events. If
-the queue is full, no new events can be inserted into the memory queue. Only
-after the signal from the output will the queue free up space for more events to be accepted.
+The memory queue waits for the output to acknowledge or drop events. If the queue is full, no new
+events can be inserted into the memory queue. Only after the signal from the output will the queue
+free up space for more events to be accepted.
 
-The memory queue is controlled by the parameters `queue.mem.flush.min_events` and `queue.mem.flush.timeout`. If
-`queue.mem.flush.timeout` is `0s` or `queue.mem.flush.min_events` is `0` or `1` then events can be sent by the output as
-soon as they are available. If the output supports a `bulk_max_size` parameter it controls the
-maximum batch size that can be sent.
+The memory queue is controlled by the parameters flush.min_events and flush.timeout.
+flush.min_events gives a limit on the number of events that can be included in a single batch, and
+flush.timeout specifies how long the queue should wait to completely fill an event request. If the
+output supports a bulk_max_size parameter, the maximum batch size will be the smaller of
+bulk_max_size and flush.min_events.
 
-If `queue.mem.flush.min_events` is greater than `1` and `queue.mem.flush.timeout` is greater than `0s`, events will only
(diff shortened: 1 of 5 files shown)


cmacknz:
The changes here are ported over from https://github.com/elastic/beats/pull/37795/files

cmacknz:
The bulk_max_size docs seem to not be up to date either here or in Beats either https://github.com/elastic/beats/blob/main/libbeat/outputs/elasticsearch/docs/elasticsearch.asciidoc#bulk_max_size

cmacknz:
Fixing the bulk_max_size docs in beats: https://github.com/elastic/beats/pull/38634

docs/en/ingest-management/elastic-agent/configuration/outputs/output-logstash.asciidoc
```diff
@@ -169,23 +169,34 @@ For more information, refer to <<secure-logstash-connections>>.
...
+For backwards compatibility, synchronous mode can also be activated by setting flush.min_events to 0
+or 1. In this case, batch size will be capped at 1/2 the queue capacity.
+
+In asynchronous mode, an event request will wait up to the specified timeout to try and fill the
+requested batch completely. If the timeout expires, the queue returns a partial batch with all
+available events. To use asynchronous mode, set flush.timeout to a positive duration, e.g. 5s.
+
+This sample configuration forwards events to the output when there are enough events to fill the
+output's request (usually controlled by bulk_max_size, and limited to at most 512 events by
+flush.min_events), or when events have been waiting for 5s without filling the requested size:f 512 events are available or the oldest
```
kilfoyle, line 199:
```suggestion
The memory queue is controlled by the parameters `flush.min_events` and `flush.timeout`.
`flush.min_events` gives a limit on the number of events that can be included in a single batch, and
`flush.timeout` specifies how long the queue should wait to completely fill an event request. If the
output supports a `bulk_max_size` parameter, the maximum batch size will be the smaller of
`bulk_max_size` and `flush.min_events`.

`flush.min_events` is a legacy parameter, and new configurations should prefer to control batch size
with `bulk_max_size`. As of 8.13, there is never a performance advantage to limiting batch size with
`flush.min_events` instead of `bulk_max_size`.

In synchronous mode, an event request is always filled as soon as events are available, even if
there are not enough events to fill the requested batch. This is useful when latency must be
minimized. To use synchronous mode, set `flush.timeout` to 0.

For backwards compatibility, synchronous mode can also be activated by setting `flush.min_events` to 0
or 1. In this case, batch size will be capped at 1/2 the queue capacity.

In asynchronous mode, an event request will wait up to the specified timeout to try and fill the
requested batch completely. If the timeout expires, the queue returns a partial batch with all
available events. To use asynchronous mode, set `flush.timeout` to a positive duration, for example 5s.

This sample configuration forwards events to the output when there are enough events to fill the
output's request (usually controlled by `bulk_max_size`, and limited to at most 512 events by
`flush.min_events`), or when events have been waiting for 5s without filling the requested size:f 512 events are available or the oldest
```

I'd recommend wrapping the parameters in backticks just for consistency.

kilfoyle (dismissed):
Just the one suggestion but otherwise LGTM!

docs/en/ingest-management/elastic-agent/configuration/outputs/output-logstash.asciidoc
cmacknz: Thanks this was lost in the copy paste I did

kilfoyle (approved):

faec (approved):

PR #986, opened 2024-03-27
Remove tech preview notice from Kafka output docs

kilfoyle:
The Kafka output type GAed in 8.13 so this removes the tech preview notice from the associated docs.


karenzone (approved):
LGTM

PR #1058, opened 2024-05-03
Clarify that Fleet Server should upgraded before other agents

kilfoyle:
This adds some text to make clear that when upgrading Elastic Agent(s) any agent acting as Fleet Server should be upgraded first.

Screenshot 2024-05-03 at 2 21 35 PM


karenzone (approved):
LGTM

PR #1069, opened 2024-05-08
Remove duplicate --delay-enroll commands from Elastic Agent install command reference

strawgate:
Remove duplicate --delay-enroll commands from Elastic Agent install command reference


kilfoyle:
@strawgate Thanks for fixing this, and I'm sorry I didn't notice your PR until now!

Separately, I actually fixed the duplicates and some mis-orderings a few days ago (via [this PR](https://github.com/elastic/ingest-docs/pull/1083)), so this one can be closed. Everything in the [command ref](https://www.elastic.co/guide/en/fleet/current/elastic-agent-cmd-options.html#elastic-agent-cmd-options) should be okay now, but if I've missed anything please let me know.

PR #1083, opened 2024-05-22
Fix duplicates and misordered parameters in commmand ref

kilfoyle:
Fixes a few small duplicates and misorderings in the command docs. I've checked through the enroll and install syntax diagram and descriptions, and these seem to be the only issues.

There might be a merge conflict for recently added options...


chriscressman (approved):
LGTM

PR #1090, opened 2024-05-28
Add note about '/liveness' endpoint for standalone agent

kilfoyle:
This updates the "[Configure monitoring for standalone Elastic Agents](https://www.elastic.co/guide/en/fleet/8.13/elastic-agent-monitoring-configuration.html)" docs page to include mention of the `/liveness` endpoint added for 8.15. Rel: https://github.com/elastic/elastic-agent/issues/107#issuecomment-2133910470

**ADD:** Please see my [comment below](https://github.com/elastic/ingest-docs/pull/1090#issuecomment-2136026119) for an additional change to the Fleet-managed agent docs.

---

Screenshot 2024-05-28 at 11 21 02 AM

### DIFF
--- a/docs/en/ingest-management/elastic-agent/configuration/elastic-agent-monitoring.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/configuration/elastic-agent-monitoring.asciidoc
@@ -38,11 +38,10 @@ To enable monitoring, set `agent.monitoring.enabled` to `true`. Also set the
 collected. If neither setting is specified, monitoring is turned off. Set
 `use_output` to specify the output to which monitoring events are sent.
 
-Setting `agent.monitoring.enabled` to `true` also enables a `/liveness` endpoint to be 
-available where that server is configured to be exposed. By default, the endpoint 
+Setting `agent.monitoring.enabled` to `true` also exposes a `/liveness` endpoint. By default, the endpoint 
 returns a `200` OK status as long as {agent}'s internal main loop is responsive and can 
 process configuration changes. It can be configured to also monitor the component states 
-to return an error if anything is degraded or failed.
+and return an error if anything is degraded or has failed.
 
 The `agent.monitoring.pprof.enabled` option controls whether the {agent} and {beats} expose the
 `/debug/pprof/` endpoints with the monitoring endpoints. It is set to `false`
--- a/docs/en/ingest-management/elastic-agent/configuration/elastic-agent-monitoring.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/configuration/elastic-agent-monitoring.asciidoc
@@ -38,6 +38,12 @@ To enable monitoring, set `agent.monitoring.enabled` to `true`. Also set the
 collected. If neither setting is specified, monitoring is turned off. Set
 `use_output` to specify the output to which monitoring events are sent.
 
+Setting `agent.monitoring.enabled` to `true` also enables a `/liveness` endpoint to be 
+available where that same server is configured to be exposed. By default, the endpoint 
+returns a `200 OK` status as long {agent}'s internal main loop is responsive and can 
+process configuration changes. It can be configured to also monitor the component states 
+to return an error if anything is degraded or failed.
+
 The `agent.monitoring.pprof.enabled` option controls whether the {agent} and {beats} expose the
 `/debug/pprof/` endpoints with the monitoring endpoints. It is set to `false`
 by default. Data produced by these endpoints can be useful for debugging but present a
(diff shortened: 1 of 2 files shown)


docs/en/ingest-management/elastic-agent/configuration/elastic-agent-monitoring.asciidoc
```diff
@@ -38,6 +38,12 @@ To enable monitoring, set `agent.monitoring.enabled` to `true`. Also set the
 collected. If neither setting is specified, monitoring is turned off. Set
 `use_output` to specify the output to which monitoring events are sent.
 
+Setting `agent.monitoring.enabled` to `true` also enables a `/liveness` endpoint to be 
+available where that server is configured to be exposed. By default, the endpoint 
+returns a `200` OK status as long as {agent}'s internal main loop is responsive and can 
+process configuration changes. It can be configured to also monitor the component states 
+to return an error if anything is degraded or failed.
```
shainaraskas, line 45: "where that server is configured to be exposed" is a little confusing to me - does that just mean at the URL for the server? Would that be different for `/debug/pprof/` and `/buffer`? if not, I'd probably skip the "where configured" bit

```suggestion
Setting `agent.monitoring.enabled` to `true` also exposes a `/liveness` endpoint. By default, the endpoint 
returns a `200` OK status as long as {agent}'s internal main loop is responsive and can 
process configuration changes. It can be configured to also monitor the component states 
and return an error if anything is degraded or has failed.
```

shainaraskas (dismissed):
approved w/ suggest!

kilfoyle:
**ADD:** The "Enable agent monitoring" setting in Fleet wasn't documented at all, so I've added it to the [Elastic Agent policies](https://www.elastic.co/guide/en/fleet/master/agent-policy.html) page that applies to Fleet-managed agents.

@cmacknz @nimarezainia   If the `/liveness` endpoint should be mentioned in the Fleet docs, maybe we should provide an example of how to call it? As Shaina mentioned, the "where that server is configured to be exposed" part might not be clear, so an example could help. 

---

Screenshot 2024-05-28 at 4 03 42 PM

karenzone (approved):
LGTM

docs/en/ingest-management/elastic-agent/configuration/elastic-agent-monitoring.asciidoc
```diff
@@ -38,6 +38,11 @@ To enable monitoring, set `agent.monitoring.enabled` to `true`. Also set the
 collected. If neither setting is specified, monitoring is turned off. Set
 `use_output` to specify the output to which monitoring events are sent.
 
+Setting `agent.monitoring.enabled` to `true` also exposes a `/liveness` endpoint. By default, the endpoint
```
cmacknz, line 41: It's actually `agent.monitoring.enabled.http` that needs to be enabled, `agent.monitoring.enabled` controls whether we collect monitoring data from agent to be shipped to Fleet.

The liveness probe is an HTTP feature. Sorry for the late review.

kilfoyle: Thanks for spotting this @cmacknz!

A couple of questions:
 - Should it be `agent.monitoring.enabled.http` or `agent.monitoring.http.enabled`? In your [comment](https://github.com/elastic/elastic-agent/issues/107#issuecomment-2133910470) you mentioned the latter.
 - I don't think the setting is available to in Fleet, so I think I should mention the setting only in the standalone agent docs ([here](https://www.elastic.co/guide/en/fleet/8.13/elastic-agent-monitoring-configuration.html)). Does that seem right to you?

zez3:
The current documentation still does not reflect the new `liveness` option

@kilfoyle can you please fix that?

nimarezainia:
@zez3 seems to be here:  https://www.elastic.co/guide/en/fleet/8.15/elastic-agent-monitoring-configuration.html

image

kilfoyle:
Thanks @nimarezainia!

zez3:
So this will work only for standalone Agents?

zez3:
>It can be configured to also monitor the component states and return an error if anything is degraded or has failed.

How?

zez3:
> So this will work only for standalone Agents?

I see https://github.com/elastic/kibana/pull/180922#issuecomment-2286182984  , so this is not yet fully functional.

cmacknz:
It isn't in the Fleet UI yet, but the Fleet overrides API will let you control it. We have [tests](https://github.com/elastic/elastic-agent/blob/44528c49507c18e71a4cfd2a6ba4f0904bd32009/testing/integration/monitoring_probe_reload_test.go#L110-L124) proving Fleet can control the parameters, there just isn't a UI box for it yet.

The request to turn on the HTTP endpoint, which also enables the /liveness endpoint would look something like the following from dev tools:

```
PUT kbn:/api/fleet/agent_policies/65df03d7-5d26-490c-a24a-ae05095c1a4a
{
    "name": "Agent policy 1",
    "namespace": "default",
    "overrides": {
      "agent": {
          "monitoring": map[string]interface{}{
            "http": map[string]interface{}{
              "enabled": true,
              "host":    "localhost",
              "port":    6792,
            },
        }
    }
}
```

There is a `failon` query parameter that controls what it considers "live" to be: https://github.com/elastic/elastic-agent/blob/44528c49507c18e71a4cfd2a6ba4f0904bd32009/elastic-agent.reference.yml#L155-L168

nimarezainia:
Fleet UI changes: https://github.com/elastic/kibana/issues/153950

PR #1147, opened 2024-07-02
Change 'enable' to 'set up' for agent alerting

kilfoyle:
Just fixing up the terminology a bit for  this section about [alerts](https://www.elastic.co/guide/en/fleet/master/monitor-elastic-agent.html#fleet-alerting). The term "enable" seems a bit inappropriate in the context of an alert, since it implies that the alert already exists.


karenzone (approved):
LGTM

Issue #1152, opened 2024-07-03
[REQUEST]: Document how to run Elastic Agent and Beats in containers with `--read-only` filesystems


ycombinator:
### Description

Typically users who run Elastic Agent or Beats in containers do so with "normal" read/write filesystems.  However, some users may want to run these products in containers with read-only filesystems.  This is achieved by passing an additional `--read-only` CLI flag to the `docker run` command.

When Elastic Agent or Beats are run in containers with a read-only filesystem, users will get errors unless they also specify a mount path to a filesystem location where the products can write their internal state.  We need to document this scenario.

### Resources

https://github.com/elastic/ingest-dev/issues/3267#issuecomment-2196796778
https://github.com/elastic/elastic-agent/pull/4995#issuecomment-2192751402

### Collaboration

The documentation team will investigate the issue and create the initial content.

### Point of contact.

**Main contact:** @blakerouse

kilfoyle:
Thanks for opening this @ycombinator!

Beats docs PR: https://github.com/elastic/beats/pull/40092
Elastic Agent docs PR: https://github.com/elastic/ingest-docs/pull/1154

kilfoyle:
I think this can be closed now:
 - Beats docs updated via https://github.com/elastic/beats/pull/40092
 - Elastic Agent docs updated via https://github.com/elastic/ingest-docs/pull/1154

PR #1154, opened 2024-07-03
Add read-only filesystem option for Elastic Agent on Docker

kilfoyle:
This updates the [Running Elastic Agent in a container](https://www.elastic.co/guide/en/fleet/current/elastic-agent-container.html#_step_4_run_the_elastic_agent_image) instructions with the options required for running the container on a read-only file system.

Target: 8.15
Closes:: https://github.com/elastic/ingest-docs/issues/1152

I assumed that "STATE_PATH is a variable that should be replaced with the full path to the stateful directory to mount, but please correct me if that's wrong.

---

Screenshot 2024-07-09 at 11 25 43 AM

### DIFF
--- a/docs/en/ingest-management/elastic-agent/elastic-agent-container.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/elastic-agent-container.asciidoc
@@ -108,7 +108,7 @@ For example:
 
 [source,terminal,subs="attributes"]
 ----
-docker run --rm --mount source=$(pwd)/state,destination=/state -e {STATE_PATH}=/state --read-only docker.elastic.co/beats/elastic-agent:8.15.0 <1>
+docker run --rm --mount source=$(pwd)/state,destination=/state -e {STATE_PATH}=/state --read-only docker.elastic.co/beats/elastic-agent:{version} <1>
 ----
 
 Where {STATE_PATH} is the path to a stateful directory to mount where {agent} application data can be stored.
--- a/docs/en/ingest-management/elastic-agent/elastic-agent-container.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/elastic-agent-container.asciidoc
@@ -95,6 +95,25 @@ docker run --rm docker.elastic.co/beats/elastic-agent:{version} elastic-agent co
 
 include::{ingest-docs-root}/docs/en/ingest-management/tab-widgets/run-agent-image/widget.asciidoc[]
 
+[TIP] 
+.Running {agent} on a read-only file system
+==== 
+If you'd like to run {agent} in a Docker container on a read-only file 
+system, you can do so by specifying the `--read-only` option.
+{agent} requires a stateful directory to store application data, so
+with the `--read-only` option you also need to use the `--mount` option to
+specify a path to where that data can be stored.
+
+For example:
+
+[source,terminal,subs="attributes"]
+----
+docker run --rm --mount type=tmpfs,destination=/state -e {STATE_PATH}=/state --read-only docker.elastic.co/beats/elastic-agent:8.15.0 <1>
+----
+
+Where {STATE_PATH} is the path to a stateful directory to mount where {agent} application data can be stored.
+====
+
 [discrete]
 == Step 5: View your data in {kib}
 
--- a/docs/en/ingest-management/elastic-agent/elastic-agent-container.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/elastic-agent-container.asciidoc
@@ -108,7 +108,7 @@ For example:
 
 [source,terminal,subs="attributes"]
 ----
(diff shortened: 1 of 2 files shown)


docs/en/ingest-management/elastic-agent/elastic-agent-container.asciidoc
```diff
@@ -95,6 +95,25 @@ docker run --rm docker.elastic.co/beats/elastic-agent:{version} elastic-agent co
...
+system, you can do so by specifying the `--read-only` option.
+{agent} requires a stateful directory to store application data, so
+with the `--read-only` option you also need to use the `--mount` option to
+specify a path to where that data can be stored.
+
+For example:
+
+[source,terminal,subs="attributes"]
+----
+docker run --rm --mount type=tmpfs,destination=/state -e {STATE_PATH}=/state --read-only docker.elastic.co/beats/elastic-agent:8.15.0 <1>
```
blakerouse, line 111: I would not recommend `type=tmpfs`. It is very much preferred that it be a real mount that can store actually data.

An optional note about the option of using a `tmpfs` could be added, but should have a warning that it can cause data duplicate upon restart of the container (mainly for logs, some metricsets are completely stateless and doesn't cause this issue) as no state information will persist.

blakerouse:
Could also add a section about doing the same in kubernetes with the `readOnlyRootFilesystem ` - https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

docs/en/ingest-management/elastic-agent/elastic-agent-container.asciidoc
kilfoyle: @blakerouse I've removed `type=tmpfs,` from the command. Let me know if I've got it wrong and instead we should specify a different value for `type`. 

I also added a warning about tmpfs directories not persisting data.

kilfoyle:
> Could also add a section about doing the same in kubernetes with the readOnlyRootFilesystem  - https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

@blakerouse  I've updated the [Kubernetes install section](http://localhost:8000/guide/running-on-kubernetes-managed-by-fleet.html#_step_5_deploy_the_elastic_agent) with the tip marked in red below. When specifying `readOnlyRootFilesystem` do we also have to specify the path of the directory to mount? If so, would you have an example of what that looks like in kubectl command?
CC @constanca-m

k8s-screen

docs/en/ingest-management/elastic-agent/elastic-agent-container.asciidoc
```diff
@@ -95,6 +95,27 @@ docker run --rm docker.elastic.co/beats/elastic-agent:{version} elastic-agent co
...
+system, you can do so by specifying the `--read-only` option.
+{agent} requires a stateful directory to store application data, so
+with the `--read-only` option you also need to use the `--mount` option to
+specify a path to where that data can be stored.
+
+For example:
+
+[source,terminal,subs="attributes"]
+----
+docker run --rm --mount destination=/state -e {STATE_PATH}=/state --read-only docker.elastic.co/beats/elastic-agent:8.15.0 <1>
```
blakerouse, line 111: A source is required if `type=tmpfs` is not used.

```suggestion
docker run --rm --mount source=$(pwd)/state,destination=/state -e {STATE_PATH}=/state --read-only docker.elastic.co/beats/elastic-agent:8.15.0 <1>
```

docs/en/ingest-management/elastic-agent/run-container-common/deploy-elastic-agent.asciidoc
```diff
@@ -17,5 +17,9 @@ elastic-agent-fj2z9             1/1     Running   0          81m
 elastic-agent-hs4pb             1/1     Running   0          81m
 ------------------------------------------------
 
-
-
+[TIP] 
+.Running {agent} on a read-only file system
+==== 
+If you'd like to run {agent} on Kubernetes on a read-only file 
+system, you can do so by specifying the `readOnlyRootFilesystem` option.
```
blakerouse, line 24: I think this is good enough, I don't know if we should really go into details on how to configure the pod specification. Any kubernetes engineer understands this option and its easily search-able to find examples on how this is done.

Optional adding it as a commented out section in our kubernetes manifests, would be a better way of doing this.

blakerouse (changes requested):

docs/en/ingest-management/elastic-agent/run-container-common/deploy-elastic-agent.asciidoc
kilfoyle: Sounds good. I've left it as is.

kilfoyle:
Thanks @blakerouse! I think everything is fixed up now.

blakerouse (approved):
Looks good.

karenzone (approved):
LGTM

PR #1226, opened 2024-08-08
[DOCS] Update link

lucabelluccini:
Please backport to all versions where the link is present.


kilfoyle (approved):
Thanks Luca! Our docs team backporting policy prevents me from backporting any earlier than the 8.14 release, but I'll make sure the change is pushed into there and all later releases.

Issue #1103, opened 2024-06-06
[Fleet]: Screenshot not updated for Integrations tab as config and API reference tab is not available.


amolnater-qasource:
**Description:**
Screenshot not updated for Integrations tab as config and API reference tab is not available.

**Link:**
https://www.elastic.co/guide/en/observability/8.14/logs-metrics-get-started.html#view-data
https://www.elastic.co/guide/en/fleet/current/upgrade-integration.html#upgrade-integration-to-latest-version

**Actual Screenshot:**
![image](https://github.com/elastic/ingest-docs/assets/77374876/13a14c2b-d0fc-4745-99c7-77ad063b42d6)

**Expected Screenshot:**
![image](https://github.com/elastic/ingest-docs/assets/77374876/7f6e96cb-c528-4008-a394-5c40fcc8eb5e)

amolnater-qasource:
@manishgupta-qasource Please review.

manishgupta-qasource:
Secondary review for this ticket is Done

amolnater-qasource:
FYI @kilfoyle

amolnater-qasource:
Hi Team,

We have validated the updated docs at links:

- https://www.elastic.co/guide/en/observability/8.17/logs-metrics-get-started.html#view-data

**Screenshots:**
![Image](https://github.com/user-attachments/assets/ae3f4bf2-1dcb-493a-903a-38dff1fa62c1)


Hence, we are marking this issue as **QA:Validated**.

Thanks!!

PR #1243, opened 2024-08-09
Fix screenshots for integrations upgrades

kilfoyle:
This updates the outdated screenshots described in https://github.com/elastic/ingest-docs/issues/1103

For the Nginx screencap, I've cropped it to just show the "New version available" warning, since it's very difficult to recreate (I'm not sure if it's possible to install an earlier version of an integration).

Closes: #1103


karenzone (approved):
LGTM

PR #1245, opened 2024-08-09
Add note to 'unprivleged' agent docs about elastic-agent-user

kilfoyle:
From a discussion in Slack, some commands when run in unprivleged mode require a `sudo -u elastic-agent-user` prefix. 

See [Knowledge Base article](https://support.elastic.dev/knowledge/view/1d161cf8) for reference.

screen

---

ADD: This also updates the command reference to mark the `elastic-agent privileged` and `elastic-agent unprivileged` as in technical preview mode, since I forgot to do that in https://github.com/elastic/ingest-docs/pull/1244

### DIFF
--- a/docs/en/ingest-management/elastic-agent/elastic-agent-unprivileged-mode.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/elastic-agent-unprivileged-mode.asciidoc
@@ -33,9 +33,16 @@ elastic-agent install \
   --unprivileged
 ----
 
-IMPORTANT: On Linux systems, once {agent} has been installed with the `--unprivileged` flag, all {agent} commands that you run should not be prefixed with `sudo`.
+[IMPORTANT] 
+====
+Note the following current restrictions for running {agent} in `unprivileged` mode:
+
+* On Linux systems, once {agent} has been installed with the `--unprivileged` flag, all {agent} commands that you run should not be prefixed with `sudo`.
 Including `sudo` in a command may result in <<agent-sudo-error,an error>> due to the agent not having the required privileges.
-The `sudo` option is still required for the `elastic-agent install` command.
+* The `sudo` option is still required for the `elastic-agent install` command.
+* Currently, when you run some {agent} commands as non-root, for example, `elastic-agent inspect`, you may need to prefix the command with `sudo -u elastic-agent-user elastic-agent`.
+For example, `sudo -u elastic-agent-user elastic-agent inspect`.
+====
 
 [discrete]
 [[unprivileged-command-behaviors]]
--- a/docs/en/ingest-management/elastic-agent/elastic-agent-unprivileged-mode.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/elastic-agent-unprivileged-mode.asciidoc
@@ -37,11 +37,20 @@ elastic-agent install \
 ====
 Note the following current restrictions for running {agent} in `unprivileged` mode:
 
-* On Linux systems, once {agent} has been installed with the `--unprivileged` flag, all {agent} commands that you run should not be prefixed with `sudo`.
-Including `sudo` in a command may result in <<agent-sudo-error,an error>> due to the agent not having the required privileges.
-* The `sudo` option is still required for the `elastic-agent install` command.
-* Currently, when you run some {agent} commands as non-root, for example, `elastic-agent inspect`, you may need to prefix the command with `sudo -u elastic-agent-user elastic-agent`.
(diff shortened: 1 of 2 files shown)


shainaraskas (approved):

docs/en/ingest-management/elastic-agent/elastic-agent-unprivileged-mode.asciidoc
```diff
@@ -33,9 +33,16 @@ elastic-agent install \
...
+====
+Note the following current restrictions for running {agent} in `unprivileged` mode:
+
+* On Linux systems, once {agent} has been installed with the `--unprivileged` flag, all {agent} commands that you run should not be prefixed with `sudo`.
 Including `sudo` in a command may result in <<agent-sudo-error,an error>> due to the agent not having the required privileges.
-The `sudo` option is still required for the `elastic-agent install` command.
+* The `sudo` option is still required for the `elastic-agent install` command.
+* Currently, when you run some {agent} commands as non-root, for example, `elastic-agent inspect`, you may need to prefix the command with `sudo -u elastic-agent-user elastic-agent`.
+For example, `sudo -u elastic-agent-user elastic-agent inspect`.
+====
```
cmacknz, line 45: We can clarify this I think. 

Running as `sudo` by default executes the command as the root user. Running as `sudo -u elastic-agent-user` runs the command as the `elastic-agent-user` that is used to run the agent service when installed with unprivileged. `sudo` is used to execute a command as another user and defaults to root. For an unprivileged agent using sudo is always necessary if the current logged in user is not `elastic-agent-user` or added to the `elastic-agent` group.

I suggest re-wording this to say something like:

* On Linux systems, once {agent} has been installed with the `--unprivileged` flag, all {agent} commands can be run without being the root user. 
  * The `sudo` option is still required for the `elastic-agent install` command. Only root can install new services. The installed service will not run as the root user.
* Using `sudo` without specifying an alternate non-root user with `sudo -u` in a command may result in <<agent-sudo-error,an error>> due to the agent not having the required privileges.
* Using `sudo -u elastic-agent-user` will execute commands as the user running the Elastic Agent service and will always work.
* For files that allow users in the `elastic-agent` group access, using an alternate user that has been added to that group will also work. There are still some commands that are only accessible to the `elastic-agent-user` that runs the service.
  * For example, `elastic-agent inspect`, requires you to prefix the command with `sudo -u elastic-agent-user`. For example, `sudo -u elastic-agent-user elastic-agent inspect`.

kilfoyle: Thanks @cmacknz. That perfect - so much more clear!

I've updated the screenshot in the description so you can see the final output.

cmacknz (approved):

docs/en/ingest-management/elastic-agent/elastic-agent-unprivileged-mode.asciidoc
cmacknz: No problem, 99% of the problems with unprivileged mode is that it forces users to have a detailed understanding of their operating system's privilege model which is not necessarily easy once you get into the fine details.

shainaraskas (approved):
one small comment - feel free to ignore!

docs/en/ingest-management/elastic-agent/elastic-agent-unprivileged-mode.asciidoc
```diff
@@ -33,9 +33,25 @@ elastic-agent install \
...
+
+* On Linux systems, after {agent} has been installed with the `--unprivileged` flag, all {agent} commands can be run without being the root user.
+** The `sudo` option is still required for the `elastic-agent install` command.
+Only `root` can install new services.
+The installed service will not run as the root user.
+* Using `sudo` without specifying an alternate non-root user with `sudo -u` in a command may result in <<agent-sudo-error,an error>> due to the agent not having the required privileges.
+* Using `sudo -u elastic-agent-user` will run commands as the user running the {agent} service and will always work.
+* For files that allow users in the `elastic-agent` group access, using an alternate user that has been added to that group will also work.
+There are still some commands that are only accessible to the `elastic-agent-user` that runs the service.
+** For example, `elastic-agent inspect` requires you to prefix the command with `sudo -u elastic-agent-user`.
```
shainaraskas, line 48: imo, examples shouldn't be included as child bullets, but instead as paragraphs in the bullet

kilfoyle: Yeah, good point! I'll leave as is for now though.

Issue #1107, opened 2024-06-06
[Fleet]: Outdated screenshots for Dashboards are visible.


amolnater-qasource:
**Description:**
Outdated screenshots for Dashboards are visible.

**Link:**
https://www.elastic.co/guide/en/fleet/current/example-standalone-monitor-nginx-serverless.html#nginx-guide-confirm-agent-data-serverless
https://www.elastic.co/guide/en/fleet/current/example-standalone-monitor-nginx.html

**Actual Screenshot:**
![image](https://github.com/elastic/ingest-docs/assets/77374876/dbd5b74b-cedb-427a-9c90-8ff197744dd9)

![image](https://github.com/elastic/ingest-docs/assets/77374876/cb018f18-6897-4173-8ddd-a569b2ed4499)

**Expected Screenshot:**
![image](https://github.com/elastic/ingest-docs/assets/77374876/8cbb6847-6dc4-4474-a7ac-450bdaf4f50b)

![image](https://github.com/elastic/ingest-docs/assets/77374876/83859b79-a1c1-47e1-922b-3b69cfc2fce2)

amolnater-qasource:
@manishgupta-qasource Please review.

manishgupta-qasource:
Secondary review for this ticket is Done

amolnater-qasource:
Hi Team,

We have validated the updated docs at links:

- https://www.elastic.co/guide/en/fleet/current/example-standalone-monitor-nginx-serverless.html#nginx-guide-confirm-agent-data-serverless
- https://www.elastic.co/guide/en/fleet/current/example-standalone-monitor-nginx.html

**Screenshots:**
![Image](https://github.com/user-attachments/assets/a4159a7e-34b4-4256-82df-6d7779980bca)
![Image](https://github.com/user-attachments/assets/bc194308-4db6-476b-9d23-2d7ae877459d)


Hence, we are marking this issue as **QA:Validated**.

Thanks!!

PR #1248, opened 2024-08-09
Fix outdated dashboard screencap in Nginx guides

kilfoyle:
This replaces the screenshots for the metrics dashboard in the two nginx tutorials, using the dashboard image from https://github.com/elastic/ingest-docs/issues/1107

I've just removed the logging dashboard screenshot for simplicity since the view no longer shows much.

Closes: #1107


shainaraskas (approved):
imo, ideally the screenshot would be trimmed to exclude the scrollbar and the tiny black line at the bottom. approving to unblock, but consider polishing it before 🚢

kilfoyle:
I've trimmed the graphic.

shainaraskas (approved):
sharp!

Issue #1222, opened 2024-08-02
[REQUEST]: Document full list of auto-updated packages 


kpollich:
### Description

Today, we include docs around Fleet's auto-update and auto-policy-upgrade logic here: https://www.elastic.co/guide/en/fleet/current/upgrade-integration.html#upgrade-integration-policies-automatically. However, we don't expose the full list of packages that are subject to this behavior by default in Fleet. We should include this list. 

### Resources

The list of packages is defined in https://github.com/elastic/kibana/blob/d4dc118ffac98ce132d9b4569fdf63ed0884383e/x-pack/plugins/fleet/common/constants/epm.ts#L70-L75 and includes

- Endpoint
- APM
- Synthetics
- Cloud Security Posture Management

I believe `system`, `elastic_agent`, and `fleet_server` are also subject to this behavior, but need to confirm by digging a bit deeper. 

### Collaboration

The documentation team will investigate the issue and create the initial content.

### Point of contact.

**Main contact:** @kpollich 

**Stakeholders:** @elastic/fleet

kilfoyle:
Thanks for opening this @kpollich. I'll wait for your confirmation on the `system`, `elastic_agent`, and `fleet_server` packages, and then it should be a quick fix.

kpollich:
Confirming the behavior above:

- Only `synthetics` is auto-installed on a fresh cluster
- `system` is installed when "Collect system logs and metrics" is enabled on an agent policy

![image](https://github.com/user-attachments/assets/d3c49814-e524-4843-9145-b75fbbd26c45)

- `elastic_agent` is installed when these agent monitoring options are enabled when creating an agent policy (they're under advanced options and enabled by default)

![image](https://github.com/user-attachments/assets/839f971d-a8a1-4d76-89fe-4fd02588a254)

- `fleet_server` is installed when setting up a Fleet Server via Fleet UI

All of these integrations can be opted into automatic policy upgrades, but it's disabled by default. So, we won't automatically upgrade your agents at the edge to the newest version of these integrations. 

![image](https://github.com/user-attachments/assets/d0ce700e-21ba-45f1-a109-07fdc12ba4d7)
![image](https://github.com/user-attachments/assets/4cac8e04-5210-48c1-abe3-cca0120eda88)
![image](https://github.com/user-attachments/assets/1797b9e5-7beb-4dff-9a6a-60d3e7352184)

[...]

PR #1260, opened 2024-08-19
Document list of auto-updated packages

kilfoyle:
This adds the list of integrations for which integration policies are updated automatically. It also makes clear for which of these you can opt-in to the auto-updates versus those for which the auto-updates can't be disabled.

Thanks @kpollich for the super clear description in the issue!

Closes: #1222

---

Screenshot 2024-08-19 at 12 04 51 PM

### DIFF
--- a/docs/en/ingest-management/integrations/upgrade-integration.asciidoc
+++ b/docs/en/ingest-management/integrations/upgrade-integration.asciidoc
@@ -63,13 +63,26 @@ and download your standalone agent policy, see <<update-standalone-policies>>.
 [[upgrade-integration-policies-automatically]]
 == Keep integration policies up to date automatically
 
-Some integration packages, like System, are installed by default during {fleet}
-setup. These integrations are upgraded automatically when {fleet} detects that a
-new version is available. These integrations have an option to upgrade
-integration policies automatically, too. When this option is selected (the
-default), {fleet} automatically upgrades your policies behind the scenes when a
-new version of the integration is available. If there are conflicts during the
-upgrade, your integration policies will not be upgraded, and you'll need to
+Some integration packages, like System, are installed by default during {fleet} setup.
+These integrations are upgraded automatically when {fleet} detects that a new version is available. 
+
+The following integrations are installed automatically when you select certain options in the {fleet} UI.
+All of them have an option to upgrade integration policies automatically, too:
+ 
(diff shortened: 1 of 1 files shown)


kpollich (approved):
Looks great! Many thanks!

docs/en/ingest-management/integrations/upgrade-integration.asciidoc
```diff
@@ -63,13 +63,26 @@ and download your standalone agent policy, see <<update-standalone-policies>>.
...
+
+The {integrations-docs}/endpoint[Elastic Defend] integration also has an option to upgrade installation policies automatically.
+
+Note that for the following integrations, when the integration is updated automatically the integration policy is upgraded automatically as well. This behavior cannot be disabled.
+
+* {integrations-docs}/apm[Elastic APM]
+* {integrations-docs}/cloud_security_posture#cloud-security-posture-management-cspm[Cloud Security Posture Management]
+* {observability-guide}/monitor-uptime-synthetics.html[Elastic Synthetics]
+
+For integrations that support the option to auto-upgrade the integration policy, when this option is selected (the default), {fleet} automatically upgrades your policies behind the scenes when a new version of the integration is available.
```
karenzone, line 84:
```suggestion
For integrations that support the option to auto-upgrade the integration policy, {fleet} automatically upgrades your policies when a new version of the integration is available.
```

karenzone (dismissed):
Left suggestion inline for consideration. Otherwise, LGTM

docs/en/ingest-management/integrations/upgrade-integration.asciidoc
kilfoyle: Thanks Karen, but I'm going to skip this one. I don't think your suggested text has the same meaning (i.e., that the auto-upgrade is enabled by default but that they're able to opt-out, unlike other integrations listed further in().

karenzone (approved):
LGTM

docs/en/ingest-management/integrations/upgrade-integration.asciidoc
kilfoyle: Thanks @karenzone. I removed the "behind the scenes" (I didn't like that part either), but left the "When the option is selected" because that's actually the important bit.

Issue #1273, opened 2024-08-30
The documentation contains a description of the non-existent environment variable `FLEET_SERVER_INSECURE_HTTP`


linghengqian:
- As we all know, on the Kibana side, registering a fleet server with the http protocol is not actually allowed. The fleet server URL must be in the https protocol.
- ![Image](https://github.com/user-attachments/assets/6ae619b5-b5e7-4073-84d4-d2efa1704837)
- In https://github.com/elastic/kibana/pull/129371 , `--fleet-server-insecure-http` has been removed.
- But in https://www.elastic.co/guide/en/fleet/current/agent-environment-variables.html#env-bootstrap-fleet-server , you can still see the explanation of `FLEET_SERVER_INSECURE_HTTP`. I think this is unreasonable.
- ![Image](https://github.com/user-attachments/assets/86d82def-3984-437f-b971-2f759f101fa5)

PR #1274, opened 2024-08-30
Remove 'fleet-server-insecure-http' setting from docs

kilfoyle:
Apparently the setting `fleet-server-insecure-http` was removed by https://github.com/elastic/kibana/pull/129371
This updates the docs to remove mention of the removed setting.

Closes: https://github.com/elastic/ingest-docs/issues/1273

### DIFF
--- a/docs/en/ingest-management/commands.asciidoc
+++ b/docs/en/ingest-management/commands.asciidoc
@@ -198,7 +198,6 @@ elastic-agent enroll --fleet-server-es <string>
                      [--fleet-server-es-cert-key <string>]
                      [--fleet-server-es-insecure]
                      [--fleet-server-host <string>]
-                     [--fleet-server-insecure-http]
                      [--fleet-server-policy <string>]
                      [--fleet-server-port <uint16>]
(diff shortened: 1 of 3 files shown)


kilfoyle:
@nchaulet  When you have a minute can you review this one, please?

nchaulet:
@kilfoyle I am not sure about that one, the UI do not allow to setup an `http` fleet server, but the option `--fleet-server-insecure-http`  still exists in fleet server and it possible to create an http fleet server hosts with the Fleet API, this is only used for testing/development.

kilfoyle:
Thanks @nchaulet! I'll update the parent issue accordingly.

kilfoyle:
Actually, if the setting is used only for testing & development, perhaps we should just remove it from the docs anyway.

nchaulet:
> Actually, if the setting is used only for testing & development, perhaps we should just remove it from the docs anyway. What do you think?

Yes it's probably better as there is no reason for an end user to use those settings

nchaulet (approved):

lcawl (approved):

Issue #1284, opened 2024-09-04
[REQUEST]: Fix agent status docs for unhealthy status


kpollich:
### Description

Currently, https://www.elastic.co/guide/en/fleet/8.15/monitor-elastic-agent.html#view-agent-status describes the unhealthy status as follows: 

> Elastic Agents have not checked in to Fleet Server. At this point, you may need to address the situation.

This is inaccurate. Agents will only report as unhealthy when one or more input/output reports as unhealthy. When agents miss check-ins for a period of time defaulting to 5 minutes, they will be considered offline. There is no intermediate state transition from healthy -> unhealthy -> offline when an agent is missing check-ins, and the agent transitions directly from healthy -> offline when it hasn't checked in for five minutes. 

Explicitly, when an agent's last check-in status value is either `error` or `degraded`, the agent's status will be reported as "unhealthy".

@cmacknz can you sanity check this to make sure I'm not misrepresenting anything? 

### Resources

- https://www.elastic.co/guide/en/fleet/8.15/monitor-elastic-agent.html#view-agent-status
- https://github.com/elastic/kibana/blob/1b2cbf15d88401249eb8017d529ce88a866ce931/x-pack/plugins/fleet/server/services/agents/build_status_runtime_field.ts#L93-L134 ("error" status is coerced to "unhealthy" [here](https://github.com/elastic/kibana/blob/e88499a27c7ae1ec668d6b4fe0e11ce9a487d0be/x-pack/plugins/fleet/server/services/agents/status.ts#L146)

### Collaboration

Please choose a preferred collaboration model.

### Point of contact.

**Main contact:** @kpollich

**Stakeholders:** @elastic/fleet @elastic/elastic-agent

blakerouse:
@kpollich Your statement is correct. Unhealthy means there is something wrong with the Elastic Agent on the host, either configuration issue with an integration, the Elastic Agent doesn't have the correct permissions for running that integration, etc.

kilfoyle:
Thanks Kyle and Blake! Here's a docs PR: https://github.com/elastic/ingest-docs/pull/1288

nimarezainia:
for reference

![Image](https://github.com/user-attachments/assets/36552a37-b671-41d0-967f-ea0acc27e612)

kilfoyle:
Amazing. Thanks a lot @nimarezainia!

PR #1288, opened 2024-09-04
Fix agent status docs for unhealthy status

kilfoyle:
This corrects the meaning of `Unhealthy` for Elastic Agent status, as documented in [View agent status overview](https://www.elastic.co/guide/en/fleet/current/monitor-elastic-agent.html#view-agent-status).

Closes: #1284

---

Screenshot 2024-09-04 at 2 21 32 PM


kpollich (commented):
The new description looks great, but I think we also need to update the diagram to match: 

image

nimarezainia:
> The new description looks great, but I think we also need to update the diagram to match:
> 
> image

yeah you are right, I'll redo this. It's a left over from before we had health reporting.

kilfoyle:
@kpollich  I've added in the nice, new graphic:

---

Screenshot 2024-09-05 at 11 25 14 AM

kpollich (approved):
Thanks so much for taking care of this, and thanks @nimarezainia for helping with the diagram ❤️

florent-leborgne (approved):
LGTM!

PR #1292, opened 2024-09-05
Add release notes for elastic/kibana#191730

jsoriano:
This change was originally not intended to be released in 8.15.1, but we are including it at the end.

https://github.com/elastic/kibana/pull/191730


kpollich (commented):
Is it worth raising this as a known issue or otherwise noteworthy change in addition to including it as a "plain" release note entry? There's some sort of risk with this new behavior, right?

jsoriano:
Added it also as a known issue.

kpollich (approved):

kilfoyle (approved):
LGTM! 🚀

Issue #1270, opened 2024-08-28
Improve enrollment token docs and terminology consistency


kilfoyle:
From @jillguyonnet in https://github.com/elastic/kibana/pull/191585#issue-2491596270:

I believe, however, that some improvements could also be brought to the documentation itself in order to fully address the above concerns. The [Fleet enrollment tokens](https://www.elastic.co/guide/en/fleet/8.15/fleet-enrollment-tokens.html) doc page already describes enrollment token creation and deletion, perhaps the following information could be added:

Why the token name must be unique.
Add an explanation around revoking tokens (marked as inactive, will be removed after expiration) with a link to the [retention period setting in Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-settings.html#api-key-service-settings-delete-retention-period).
A mention that inactive tokens are still returned by the GET API and visible in the UI until they are cleaned up.
A final note around the variations in terminology (not a big issue, I think):

"enrollment tokens" in the docs vs. "enrollment API keys" in the OpenAPI spec (I thought it would be best to keep the latter consistent)
"[revoke token](https://www.elastic.co/guide/en/fleet/8.15/fleet-enrollment-tokens.html#revoke-fleet-enrollment-tokens)" vs. "[invalidate API key](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-invalidate-api-key.html)"

---

For the terminology, the docs need to match the UI, so perhaps the best thing would be just to add a note on the [Fleet enrollment tokens](https://www.elastic.co/guide/en/fleet/8.15/fleet-enrollment-tokens.html) doc page explaining that enrollment token = enrollment API key. 

For some other details see this comment: https://github.com/elastic/kibana/issues/155550#issuecomment-2314866970

PR #1295, opened 2024-09-05
Fix up enrollment token docs

kilfoyle:
This adds some detail to clear up usage of Fleet enrollment tokens.

Thanks for the suggestions @jillguyonnet. Let me know if I've missed anything.

Closes: #1270 

---

> Why the token name must be unique.

I think it should be enough for users to know just that the names need to be unique. I added:

Screenshot 2024-09-05 at 1 26 29 PM

---

> Add an explanation around revoking tokens (marked as inactive, will be removed after expiration) with a link to the [retention period setting in Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-settings.html#api-key-service-settings-delete-retention-period).
> A mention that inactive tokens are still returned by the GET API and visible in the UI until they are cleaned up.

Screenshot 2024-09-05 at 1 41 12 PM

---

> "enrollment tokens" in the docs vs. "enrollment API keys" in the OpenAPI spec

Screenshot 2024-09-05 at 1 31 33 PM

---

> "[revoke token](https://www.elastic.co/guide/en/fleet/8.15/fleet-enrollment-tokens.html#revoke-fleet-enrollment-tokens)" vs. "[invalidate API key](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-invalidate-api-key.html)"

Screenshot 2024-09-05 at 1 33 34 PM

### DIFF
--- a/docs/en/ingest-management/security/enrollment-tokens.asciidoc
+++ b/docs/en/ingest-management/security/enrollment-tokens.asciidoc
@@ -84,7 +84,7 @@ To re-enroll your {agent}s, use an active enrollment token.
 
 Note that when an enrollment token is revoked it is not immediately deleted.
 Deletion occurs automatically after the duration specified in the {es}
-{ref}/security-settings.html#api-key-service-settings-delete-retention-period[`xpack.security.authc.api_key.delete.retention_period`] setting has expired.
+{ref}/security-settings.html#api-key-service-settings-delete-retention-period[`xpack.security.authc.api_key.delete.retention_period`] setting has expired (see {ref}/security-api-invalidate-api-key.html[Invalidate API key API] for details).
 
 Until the enrollment token has been deleted:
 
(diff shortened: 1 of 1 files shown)


jillguyonnet (dismissed):
Looks great, thank you 🚀 

If I was to be super nitpicky, I could suggest adding a link to https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-invalidate-api-key.html, for example:

> ...setting has expired (see [Invalidate API key API](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-invalidate-api-key.html) for details).

I agree with you regarding only mentioning that the token name must be unique. The reason behind it is not 100% clear and there is a possibility it will not always be the case. I've opened https://github.com/elastic/kibana/issues/192469 to capture the open discussion that resulted from the previous doc improvement.

kilfoyle:
Good idea @jillguyonnet! I've added that link:

Screenshot 2024-09-10 at 10 00 39 AM

kilfoyle:
P.S., the new commit automatically removed your approval. Sorry. :-)

jillguyonnet (approved):

jillguyonnet:
Reapproved 🙂  Thanks for the change!

karenzone (approved):
LGTM

Issue #287, opened 2023-01-11
[REQUEST]: Document air-gapped list of architectures / package types / artifact names


lucabelluccini:
### Description

The page https://www.elastic.co/guide/en/fleet/current/air-gapped.html#air-gapped-diy-epr shows the example of artifacts only for `linux`.
It would be nice to mention the possible values for `<artifact_name>-<version>-<arch>-<package_type>` or even develop a short script to pull all the values from EPR or Artifact API and generate the command to be ran.

### Resources

Reach out privately as those are questions shared internally

### Collaboration

TBD. The docs and product team will work together to determine the best path forward.

### Point of contact.

**Main contact:** @lucabelluccini @111andre111 

**Stakeholders:**

lucabelluccini:
Examples:
```
https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.15.0-linux-arm64.tar.gz
https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.15.0-linux-x86_64.tar.gz
https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.15.0-windows-x86_64.zip
https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.15.0-darwin-x86_64.tar.gz
https://artifacts.elastic.co/downloads/beats/elastic-agent/elastic-agent-8.15.0-darwin-aarch64.tar.gz
```

The artifacts could be derived from https://www.elastic.co/api/product_versions (but we're missing some of them.
Alternatively, it could be just be used to get the versions.

nimarezainia:
@lucabelluccini is the ask here to document/pull all the assets in the EPR?
If you could elaborate the use case or what the customer was trying to achieve that this would help with. If it's the content of the EPR, I wonder if there's a better way of doing it rather than documentation.

is this the list you are referring to: https://www.elastic.co/guide/en/fleet/current/air-gapped.html#host-artifact-registry

lucabelluccini:
Sorry Nima I have to admit the issue description is not clear.

The ask is: Artifact Repository (not EPR) docs contain only assets for Linux, while we need a sort of script to to generate the content so that people can automatically pull assets for all platforms and architectures.

Let me see if I can contribute with a script. I'll update the issue within today.

lucabelluccini:
I've written a very minimal Go program to download stuff from our https://www.elastic.co/api/product_versions
[...]

PR #1305, opened 2024-09-09
Document format for air-gapped downloadable packages

kilfoyle:
This updates section 1b) of [Host your own artifact registry for binary downloads](https://www.elastic.co/guide/en/fleet/current/air-gapped.html#host-artifact-registry) to expand on the package URL format. 

it's not perfect but I don't think we have a way of listing all possible package URLs, so at least providing the format and guiding people to refer to the Elastic Agent download page can help clear things up.

Closes: https://github.com/elastic/ingest-docs/issues/287

DOCS PREVIEW

---

Screenshot 2024-09-09 at 3 58 59 PM


karenzone (approved):
LGTM

PR #1312, opened 2024-09-11
Add known issue that Beats do not recover from transient network errors

cmacknz:
See https://github.com/elastic/beats/issues/40705

- Relates https://github.com/elastic/beats/pull/40767


kilfoyle (approved):
LGTM! 👍 
Thanks for looking after this @cmacknz!

kilfoyle:
P.S. I added a `backport-8.15` so this will go into the 8.15 docs.

Issue #1314, opened 2024-09-12
[REQUEST]: Remove all mention of Beats namespace for Elastic Agent docker images


jlind23:
### Description

Elastic Agent docker images are available under two different namespaces:
- `elastic-agent`
- `beats`

Our document almost always use the `beats` and we should update it to rather talk exclusively about the `elastic-agent` one.
Example: https://www.elastic.co/guide/en/fleet/current/elastic-agent-container.html

From 9.0 onwards, we plan to remove Elastic Agent images from the `beats` namespace.

This change can be performed in 8.15.x without issue.

### Resources

Example: https://www.elastic.co/guide/en/fleet/current/elastic-agent-container.html

### Collaboration

The documentation team will investigate the issue and create the initial content.

### Point of contact.

**Main contact:** @jlind23 

**Stakeholders:** @ycombinator @cmacknz @pierrehilbert @nimarezainia

kilfoyle:
Here are a couple of PRs to get the docs fixed up:
 - Fleet & Agent docs: https://github.com/elastic/ingest-docs/pull/1315
 - Observability docs: https://github.com/elastic/observability-docs/pull/4246

PR #1315, opened 2024-09-12
[Fleet & Agent docs] Change 'beats' namespace to 'elastic-agent'

kilfoyle:
Elastic Agent Docker images are currently available under both a `beats` and `elastic-agent` namespace. For 9.0 only the latter will be available, so in preparation for that this will update the current (8.15).

Closes: https://github.com/elastic/ingest-docs/issues/1314

### DIFF
--- a/docs/en/ingest-management/elastic-agent/advanced-kubernetes-managed-by-fleet.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/advanced-kubernetes-managed-by-fleet.asciidoc
@@ -73,12 +73,12 @@ Inside the downloaded manifest, update the Daemonset resource:
 ------------------------------------------------
 containers:
   - name: elastic-agent
-    image: docker.elastic.co/beats/elastic-agent: <ImageVersion>     
+    image: docker.elastic.co/elastic-agent/elastic-agent: <ImageVersion>     
     args: ["-c", "/etc/elastic-agent/agent.yml", "-e"]
 ------------------------------------------------
 
 Notes:: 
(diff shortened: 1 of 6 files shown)


jlind23 (approved):

karenzone (approved):
LGTM

cmacknz (approved):
Doing a search for `/beats` with `rg '/beats'` shows me there might be a link that was missed.

```
docs/en/ingest-management/elastic-agent/elastic-agent-container.asciidoc
4:You can run {agent} inside a container -- either with {fleet-server} or standalone. Docker images for all versions of {agent} are available from the https://www.docker.elastic.co/r/beats/elastic-agent[Elastic Docker registry]. If you are running in Kubernetes, refer to {eck-ref}/k8s-elastic-agent.html[run {agent} on ECK].
```

However `https://www.docker.elastic.co/r/beats/elastic-agent` exists and `https://www.docker.elastic.co/r/elastic-agent/elastic-agent does not. I will follow up to investigate that, can't change it until there is a working alternative.

kilfoyle:
Thanks for catching that @cmacknz. I'll merge this and will be happy to fix up the link when we have an alternative.

cmacknz:
https://www.docker.elastic.co/r/elastic-agent/elastic-agent exists now so we can make the change in https://github.com/elastic/ingest-docs/pull/1315#pullrequestreview-2303299330

kilfoyle:
> https://www.docker.elastic.co/r/elastic-agent/elastic-agent exists now so we can make the change in https://github.com/elastic/ingest-docs/pull/1315#pullrequestreview-2303299330

Nice! Here's a PR: https://www.docker.elastic.co/r/elastic-agent/elastic-agent exists now so we can make the change in https://github.com/elastic/ingest-docs/pull/1315#pullrequestreview-2303299330

PR #1349, opened 2024-09-26
Clarification of MSI Upgrading Behavior

tyoungs-estc:
Rephrased to clarify that upgrades via fleet are supported, but the impact is that the MSI version remains registered at the initially installed package version. Added suggestion for MSI installs (package management) that in order to upgrade an uninstall/reinstall is required.


kilfoyle (approved):
LGTM! 🚀 
Thanks for making this change @tyoungs-estc!
I added a couple of labels so that the change will backport into the current 8.15 docs, but if that shouldn't be the case feel free to remove them.

PR #1350, opened 2024-09-26
Add note about elliptic-curve restriction

kilfoyle:
This adds a note at the bottom of the Overview section on the [Elastic Agent deployment models with mutual TLS](https://www.elastic.co/guide/en/fleet/current/mutual-tls.html) page and the Prerequisites section of the [Configure SSL/TLS for self-managed Fleet Servers](https://www.elastic.co/guide/en/fleet/current/secure-connections.html#prereqs) page.

Screenshot 2024-09-27 at 11 12 56 AM

@AndersonQ, please fix this up however it should be worded because I have no idea what I'm talking about here. :-)

### DIFF
--- a/docs/en/ingest-management/security/mutual-tls.asciidoc
+++ b/docs/en/ingest-management/security/mutual-tls.asciidoc
@@ -48,6 +48,8 @@ When mTLS is required, the secure setup between {agent}, {fleet}, and {fleet-ser
 .. If the {agent} policy contains mTLS configuration settings, those settings will take precedence over those used during enrollment: This includes both the mTLS settings used for connectivity between {agent} and {fleet-server} (and the {fleet} application in {kib}, for {fleet}-managed {agent}), and the settings used between {agent} and it's specified output.
 .. If the {agent} policy does not contain any TLS, mTLS, or proxy configuration settings, these settings will remain as they were specified when {agent} enrolled. Note that the initial TLS, mTLS, or proxy configuration settings can not be removed through the {agent} policy; they can only be updated.
 
+IMPORTANT: When you run {agent} with the {elastic-defend} integration, configuring an mTLS connection between {agent} and {fleet-server} does not support using an link:https://en.wikipedia.org/wiki/Elliptic-curve_cryptography[elliptic-curve (ECC) key].
+
 [discrete]
 [[mutual-tls-on-premise]]
 == On-premise deployments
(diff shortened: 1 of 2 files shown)


docs/en/ingest-management/security/mutual-tls.asciidoc
```diff
@@ -48,6 +48,8 @@ When mTLS is required, the secure setup between {agent}, {fleet}, and {fleet-ser
 .. If the {agent} policy contains mTLS configuration settings, those settings will take precedence over those used during enrollment: This includes both the mTLS settings used for connectivity between {agent} and {fleet-server} (and the {fleet} application in {kib}, for {fleet}-managed {agent}), and the settings used between {agent} and it's specified output.
 .. If the {agent} policy does not contain any TLS, mTLS, or proxy configuration settings, these settings will remain as they were specified when {agent} enrolled. Note that the initial TLS, mTLS, or proxy configuration settings can not be removed through the {agent} policy; they can only be updated.
 
+IMPORTANT: When you run {agent} with the {elastic-defend} integration, configuring an mTLS connection between {agent} and {fleet-server} does not support using an link:https://en.wikipedia.org/wiki/Elliptic-curve_cryptography[elliptic-curve (ECC) key].
```
AndersonQ, line 51: let me double check one thing with the defend folks. I believe it does not support elliptic-curve for TLS as a whole. That means for TLS and mTLS between defend and fleet-server or elasticsearch. Therefore it should also be on the TLS doc.

I'll get back to you as soon as I have an answer.

kilfoyle: Awesome. Thanks Anderson!

AndersonQ: I confirmed, the restriction is for all TLS certificate, CA or certificate key.

Better to go the other way around and say what defend accepts rather than what it does not, that means only RSA should be used.

so let me explain it. Elastic Defend connects to fleet-server and elasticsearch. Both connections normally use TLS and they might use mTLS as well. Either way the [TLS certificates](https://en.wikipedia.org/wiki/X.509) can only generated using [RSA](https://en.wikipedia.org/wiki/RSA_(cryptosystem)). The goos thing it's the most common algorithm used for TLS certificates, that's probably why no one have complained about that :)

I'll abstain from trying to write how it should be phrased, I believe you'd a way better job than I. 
Let me know if it's clear enough and I'll be happy to do the technical check on the final version.

AndersonQ: [here](https://www.elastic.co/guide/en/fleet/current/elastic-agent-ssl-configuration.html) we have the list of supported algorithms for the agent/fleet.
I don't think it's necessary to list them when explaining defend only supports RSA, but it might be good to link to that page.

also, those are valid for either the stand-alone or fleet-managed

AndersonQ (approved):

docs/en/ingest-management/security/mutual-tls.asciidoc
AndersonQ: awesome!

leemthompo (approved):
🚗

Issue #1333, opened 2024-09-19
Docs for advanced agent monitoring options UI


kilfoyle:
**From Jen in [comment](https://github.com/elastic/kibana/pull/193361#discussion_r1765736079):**

---

@kilfoyle We need to consolidate or duplicate the information on these two pages somehow:

https://www.elastic.co/guide/en/fleet/current/elastic-agent-monitoring-configuration.html

https://www.elastic.co/guide/en/fleet/current/agent-policy.html#change-policy-enable-agent-monitoring

With this PR, all the options from the first page are now available to be configured in the UI.

Based on the original issue description, https://github.com/elastic/kibana/issues/153950#issue-1645981330, it seems the first page is not capturing all the available options under agent.monitoring either.

When this is done we also need to remove this section in favor of whatever documentation is added about the full advance monitoring options:
https://www.elastic.co/guide/en/fleet/current/agent-policy.html#agent-policy-http-monitoring

---

Planned for 8.16
Jen will need an updated URL for the UI link.

PR #1361, opened 2024-10-01
Add docs for advanced monitoring options

kilfoyle:
This updates the [Enable agent monitoring](https://www.elastic.co/guide/en/fleet/current/agent-policy.html#change-policy-enable-agent-monitoring) section of the docs, renamed to "Configure agent monitoring", and adds the advanced monitoring options shown below.

This also removes the [Override the default monitoring port](https://www.elastic.co/guide/en/fleet/current/agent-policy.html#agent-policy-http-monitoring) section since that's replaced in the UI by the new monitoring options.

target: 8.16
closes: https://github.com/elastic/ingest-docs/issues/1333

I've also opened https://github.com/elastic/ingest-docs/issues/1362 as a separate issue to add a few missing options to the [Configure monitoring for standalone Elastic Agents](https://www.elastic.co/guide/en/fleet/current/elastic-agent-monitoring-configuration.html) page.

@cmacknz and @jen-huang  It'd be nice to provide some additional detail for the `/liveness` endpoint if we can. For example, how to use it in Kubernetes to restart the container. Would you know where I can find that, or if there's something I can add a link to?

---

**Advanced agent monitoring settings docs**

---

Screenshot 2024-10-02 at 12 29 34 PM

---

**Command reference docs**

---

commandref

### DIFF
--- a/docs/en/ingest-management/agent-policies.asciidoc
+++ b/docs/en/ingest-management/agent-policies.asciidoc
@@ -116,10 +116,6 @@ The following table illustrates the {fleet} user actions available to different 
 |{y}
 |{n}
 
-|<<agent-policy-http-monitoring>>
-|{y}
-|{n}
-
 |<<agent-policy-log-level>>
 |{y}
 |{n}
@@ -434,22 +430,6 @@ Select the name of the policy you want to edit.
 . Click the **Settings** tab and scroll to **Advanced settings**.
 
 . Set **Limit CPU usage** as needed. For example, to limit Go processes supervised by {agent} to two operating system threads each, set this value to `2`.
-
-[discrete]
-[[agent-policy-http-monitoring]]
-== Override the default monitoring port
-
-You can override the default port that {agent} uses to send monitoring data. It's useful to be able to adjust this setting if you have an application running on the machine on which the agent is deployed, and that is using the same port.
-
-. In {fleet}, click **Agent policies**.
-Select the name of the policy you want to edit.
-
-. Click the **Settings** tab and scroll to **Advanced settings**.
-
-//. Set **Agent HTTP monitoring** setting to enabled, and then specify a host and port for the monitoring data output.
-. Specify a host and port for the monitoring data output.
-
-//. Enable **buffer.enabled** if you'd like {agent} and {beats} to collect metrics into an in-memory buffer and expose these through a `/buffer` endpoint. This data can be useful for debugging or if the {agent} has issues communicating with {es}. Enabling this option may slightly increase process memory usage.
 
 [discrete]
 [[agent-policy-log-level]]
--- a/docs/en/ingest-management/agent-policies.asciidoc
+++ b/docs/en/ingest-management/agent-policies.asciidoc
@@ -96,7 +96,7 @@ The following table illustrates the {fleet} user actions available to different 
 |{y}
 |{n}
 
-|<<change-policy-enable-agent-monitoring,Enable agent monitoring>>
+|<<change-policy-enable-agent-monitoring,Configure agent monitoring>>
 |{y}
 |{n}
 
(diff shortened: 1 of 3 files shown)


jen-huang (dismissed):
Docs LGTM! Sorry I can't help with the Kubernetes example :(

docs/en/ingest-management/agent-policies.asciidoc
```diff
@@ -310,19 +306,40 @@ Note that adding custom tags is not supported for a small set of inputs:
...
+. Expand the **Advanced monitoring options** section to access <<advanced-agent-monitoring-settings,other monitoring settings>>.
+
+. Save your changes for the updated monitoring settings to take effect.
+
+[discrete]
+[[advanced-agent-monitoring-settings]]
+=== Advanced agent monitoring settings
+
+**HTTP monitoring endpoint**::
+Enabling this setting exposes a `/liveness` API endpoint that you can use to monitor {agent} health. By default, the endpoint returns a `200` OK status as long as {agent}'s internal main loop is responsive and can process configuration changes. It can be configured to also monitor the component states and return an error if anything is degraded or has failed. This endpoint can be used by Kubernetes to restart the container, for example.
```
cmacknz, line 331: I would include the full detail in the reference YAML to save people a trip there.

https://github.com/elastic/elastic-agent/blob/95021e1287a2a637df9c84c0e4bff0bfc661f90c/elastic-agent.reference.yml#L155-L168

docs/en/ingest-management/agent-policies.asciidoc
```diff
@@ -310,19 +306,40 @@ Note that adding custom tags is not supported for a small set of inputs:
...
++
+When you enable this setting, you need to also provide the host URL and port where the endpoint can be accessed. Using the default `localhost` is recommended.
++
+You can also enable profiling at `/debug/pprof` to control whether the {agent} exposes the `/debug/pprof/` endpoints with the monitoring endpoints. This is disabled by default. Data produced by these endpoints can be useful for debugging but present a security risk. It's recommended to leave this option disabled if the monitoring endpoint is accessible over a network.
```
cmacknz, line 335: You can mention that the heap profiles available from `/debug/pprof/` are included in diagnostics by default. CPU profiles are also included when the `--cpu-profile` option is included. On that note the CPU profile option is missing from https://www.elastic.co/guide/en/fleet/current/elastic-agent-cmd-options.html#elastic-agent-diagnostics-command.

```
❯ ./elastic-agent diagnostics --help
This command gathers diagnostics information from the Elastic Agent and writes it to a zip archive.

Usage:
  elastic-agent diagnostics [flags]

Flags:
  -p, --cpu-profile      wait to collect a CPU profile
```

The profiles exposed by `/debug/pprof` are documented in https://pkg.go.dev/net/http/pprof if you want to link to that.

docs/en/ingest-management/agent-policies.asciidoc
```diff
@@ -310,19 +306,40 @@ Note that adding custom tags is not supported for a small set of inputs:
...
+
+**Diagnostics rate limiting**::
+You can set a rate limit for the request diagnostics action handler. By default requests are limited to an interval of `1m` and a burst value of `1`. This setting does not affect diagnostics collected through the CLI.
+
+**Diagnostics file upload**::
+This setting configures retries for the file upload client. By default, a maximum of `10` retries are allowed with an initial duration of `1s` and a backoff duration of `1m`. The client may retry failed requests with exponential backoff.
```
cmacknz, line 341: You may want to mention this only have an effect for Fleet managed agents and they apply to diagnostics requests from Fleet.

cmacknz:
For using the liveness endpoint with kubernetes, it is directly usable as a liveness probe which is a first class concept in kuberentes: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-http-request

You could just link to that documentation, or we could provide a snippet that works for one (though I am way of including it because we don't test the docs so it could fall out of date).

docs/en/ingest-management/agent-policies.asciidoc
kilfoyle: I've updated the description to include the content from the reference yaml.

docs/en/ingest-management/agent-policies.asciidoc
kilfoyle: Perfect. I've added a link to the pprof docs and also updated the Command Ref to show both the `--cpu-profile` option and the `--p` option as an alias for `--cpu-profile`.

docs/en/ingest-management/agent-policies.asciidoc
kilfoyle: @cmacknz I opened a [separate issue](https://github.com/elastic/ingest-docs/issues/1362) to add these two options to the [Configure monitoring for standalone Elastic Agents](https://www.elastic.co/guide/en/fleet/current/elastic-agent-monitoring-configuration.html) page, but now I'm thinking that the settings shouldn't be documented there since they affect only Fleet-managed agent. If you agree, I'll just close that issue.

kilfoyle:
> You could just link to that documentation, or we could provide a snippet that works for one (though I am way of including it because we don't test the docs so it could fall out of date).

Perfect. I've added a link to those Kubernetes docs.

kilfoyle:
I've addressed comments and also updated the PR description to show how the latest version will render.

docs/en/ingest-management/agent-policies.asciidoc
cmacknz: Yes, no reason to document these for standalone agents as they only apply to fleet managed agents.

cmacknz (approved):

karenzone (approved):
LGTM 🚀

Issue #1395, opened 2024-10-17
[REQUEST]: Discourage cloning index templates - offer easier option to customize settings if not using namespaces + Warning/Disclaimer page


lucabelluccini:
### Description

We're aware of the issues cloning index templates of Fleet integration leads to.

We need to attempt to warn users / push users to not use index template cloning or edit the index & component templates.

The page https://www.elastic.co/guide/en/fleet/current/data-streams-ilm-tutorial.html should be reviewed to display 2 options:
- one in case the user wants to edit the ILM policy (this would involve creating a `@custom` component template to override the ILM policy or edit the default ILM policy coming with the integration) - point to https://www.elastic.co/guide/en/fleet/current/data-streams.html#data-streams-ilm
- one in case the user wants to have a namespace specific ILM policy (this will require index template cloning)
    - for the namespace specific customizations, we need to point to a new "warnings/disclaimer" page 

Further enhance https://www.elastic.co/guide/en/fleet/current/integrations-assets-best-practices.html

- **Never edit managed assets of Fleet** (e.g. the Index template, the `@package` component templates and/or ingest pipelines bundled with integrations). Any change will be overwritten upon upgrade/install.

- **When the user creates a `@custom` component template for a package integration**, while allowed and legit, the user can incur in some risks which might prevent the correct ingestion of data.
    - It can lead to broken indexing, data loss, broken integration packages upgrades.
        - Example: the `@package` component template of the integration switches from a "normal" data stream to TSDB or LogsDB and some settings/mappings introduced by the user are not compatible with such indexing modes.
        - Example: the user overrides the type of an ECS field from `keyword` to `text` - this might prevent the aggregation on such field on built-in dashboards.
        - We warn about this on [this page](https://www.elastic.co/guide/en/fleet/current/data-streams.html#:~:text=Custom%20index%20mappings%20may%20conflict%20with%20the%20mappings%20defined%20by%20the%20integration%20and%20may%20break%20the%20integration%20in%20Kibana.%20Do%20not%20change%20or%20customize%20any%20default%20mappings.)

- **When users create a `@custom` ingest pipeline** as we document [here](https://www.elastic.co/guide/en/fleet/current/data-streams-pipeline-tutorial.html), Elastic is not responsible of the correct indexing/behavior of the ingest pipeline. It becomes a custom processing of the data and it's user responsibility to test it.
[...]

PR #1408, opened 2024-10-24
Add warnings about cloning index templates

kilfoyle:
As part of https://github.com/elastic/ingest-docs/issues/1395, this updates the [Best practices for integration assets](https://www.elastic.co/guide/en/fleet/current/integrations-assets-best-practices.html) page to add new warnings.

@lucabelluccini  one question: i've added the warning about "Cloning the index template of an integration package" as suggested in the issue, but isn't this kind of repeating what's already covered in the warning about [Copying Fleet and Elastic Agent integration assets](https://www.elastic.co/guide/en/fleet/current/integrations-assets-best-practices.html#assets-restrictions-copying)?

Anyhow, please let me know whatever changes we might need. 

Closes: #1395 

---

Screenshot 2024-10-24 at 4 10 43 PM

### DIFF
--- a/docs/en/ingest-management/data-streams.asciidoc
+++ b/docs/en/ingest-management/data-streams.asciidoc
@@ -103,6 +103,7 @@ For data streams, the index template configures the stream's backing indices as 
 These templates are loaded when the integration is installed, and are used to configure the integration's data streams.
 
 [discrete]
+[[data-streams-index-templates-edit]]
 == Edit the {es} index template
 
 WARNING: Custom index mappings may conflict with the mappings defined by the integration
@@ -186,6 +187,8 @@ If left uncreated, this pipeline has no effect on your data. However, if added t
 this pipeline can be used for custom data processing, adding fields, sanitizing data, and more.
 
 Starting in version 8.12, ingest pipelines can be configured to process events at various levels of customization.
+
+NOTE: If you create a custom index pipeline, Elastic is not responsible for ensuring that it indexes and behaves as expected. Creating a custom pipeline involves custom processing of the incoming data, which should be done with caution and tested carefully.
 
 `global@custom`::
 Apply processing to all events
@@ -305,6 +308,7 @@ Specifically, apply the built-in `90-days-default` {ilm-init} policy so that dat
 
 NOTE: If you cloned an index template to customize the data retention policy on an {es} version prior to 8.13, you must update the index template in the clone to use the `ecs@mappings` component template on {es} version 8.13 or later. See <<data-streams-pipeline-update-cloned-template-before-8.13,Update index template cloned before {es} 8.13>> for the step-by-step instructions.
 
+NOTE: 
 
 [discrete]
 [[data-streams-ilm-one]]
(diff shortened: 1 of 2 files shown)


kilfoyle:

lucabelluccini:
> @lucabelluccini one question: i've added the warning about "Cloning the index template of an integration package" as suggested in the issue, but isn't this kind of repeating what's already covered in the warning about [Copying Fleet and Elastic Agent integration assets](https://www.elastic.co/guide/en/fleet/current/integrations-assets-best-practices.html#assets-restrictions-copying)?

No, it's not the same thing - let me expand.

A very common task for users is to customize the ILM policy.

Customizing the ILM policy for integration data streams should be done just using the `@custom` component template of a specific data stream.
Example: I want `metrics-elastic_agent.metricbeat*` data streams to have an ILM policy which deletes data after 1 day?
I create `metrics-elastic_agent.metricbeat@custom` component template, I add the ILM policy in its setting and I rollover. The `metrics-elastic_agent.metricbeat@custom` will be picked up automatically by the default Index Template of the integration.

Unfortunately, we dedicate a whole page to how to customize the ILM policy of a **namespace specific** data stream ([here](https://www.elastic.co/guide/en/fleet/current/data-streams-ilm-tutorial.html)).
This is an even more power-user use-case, where the customer uses separate namespaces for their data streams.
But it seems users see this as the "default" way of customizing ILM policies.

The technique of cloning the index templates to make them namespace specific is known to lead to possible problems on stack upgrades which introduce _major changes_ to the index templates.
We have a major issue on ES side on how to make those customizations without the need of cloning index templates https://github.com/elastic/elasticsearch/issues/97664#issuecomment-2413443008.

So my recommendation with this doc PR is:
- If the user just wants to change the ILM Policy or the number of shards (or other stuff) for the datastreams of one or more integrations, but they do not need to be namespace specific, then we should promote the use of the `<package>@custom` component template. This usually is fine and doesn't interfere in upgrades (except if users start to use it for heavy customizations). This should be the PROMINENT/MOST ADVERTISED option as it's likely to generate less problems to users.
- If the user wants to change the ILM Policy or the number of shards (or other stuff) for the datastreams of one or more integrations, but **they really need to be namespace specific**, then we can recommend what we say in [this doc page](https://www.elastic.co/guide/en/fleet/current/data-streams-ilm-tutorial.html).

lucabelluccini:
I've read the changes you've done @kilfoyle and they seem excellent.

I would rework the part we've mentioned in https://github.com/elastic/ingest-docs/pull/1408#issuecomment-2438151779

kilfoyle:
Thanks for the explanation @lucabelluccini! Sorry it took me a while to circle back to this PR.

 - On the [Best practices for integration assets](https://www.elastic.co/guide/en/fleet/current/integrations-assets-best-practices.html) page, I've expanded the warning about "Cloning the index template of an integration package":

screen2

---

 - In the "[Customize data retention policies](https://www.elastic.co/guide/en/fleet/current/data-streams-ilm-tutorial.html)" tutorial, the cloning happens in Step 3, so there I've added in similar guidance with the two different options you mentioned:

screen

kilfoyle:
Hi @lucabelluccini  When you have a few minutes, please let me know if the latest changes (in red above) look okay. Thanks!

lucabelluccini:
Hello @kilfoyle 

> In the "[Customize data retention policies](https://www.elastic.co/guide/en/fleet/current/data-streams-ilm-tutorial.html)" tutorial, the cloning happens in Step 3, so there I've added in similar guidance with the two different options you mentioned:

IMHO it is still a trap. This cannot be something in a info box.

**To just customize the ILM policies of a data stream regardless of a namespace (which is likely 95% of cases)**, it would be enough to just follow the step 2 of https://www.elastic.co/guide/en/fleet/current/data-streams-ilm-tutorial.html#data-streams-ilm-two

Instead, we're "suggesting" customers to clone the index template and then create a component template, which is only necessary if someone wants to customize something at a namespace level.

The [page](https://www.elastic.co/guide/en/fleet/current/data-streams-ilm-tutorial.html) should be split into 2 separate sections:
- How to customize ILM policy for a data stream belonging to an integration: simple case, 95% of situations, just create/edit the `<package>@custom` component template and add the ILM policy.
- How to customize ILM policies for a data stream belonging to an integration for a specific namespace: advanced case, 5% of situations, steps as they're described now

kilfoyle:
Closing this one in favour of the newer https://github.com/elastic/ingest-docs/pull/1623

PR #1414, opened 2024-10-31
Update the elastic-agent resource consumption docs

alexsapran:
Closes: https://github.com/elastic/ingest-dev/issues/4366

### DIFF
--- a/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
@@ -89,7 +89,7 @@ We tested using an AWS `m7i.xlarge` instance type with 4 vCPUs, 16.0 GB of memor
 | **Read network throughput** | ~40 KB/s | ~2.4 KB/s
 |===
 
-\* including all monitoring processes
+^*^ including all monitoring processes
 
 Adding integrations will increase the memory used by the agent and its processes.
 
--- a/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
@@ -78,7 +78,7 @@ Using our lab environment as an example, we can observe the following resource c
 [discrete]
 === CPU and RSS memory size
 
-// lint ignore 2 vCPU 16.0 GiB
+// lint ignore 2 vCPU 8.0 GiB
 We tested using an AWS `m7i.large` instance type with 2 vCPUs, 8.0 GB of memory, and up to 12.5 Gbps of bandwidth. The tests ingested a single log file using both the <<output-elasticsearch-performance-tuning-settings,throughput and scale preset>> with self monitoring enabled.
 These tests are representative of use cases that attempt to ingest data as fast as possible. This does not represent the resource overhead when using {integrations-docs}/endpoint[{elastic-defend}].
 [options,header]
--- a/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
@@ -79,7 +79,7 @@ Using our lab environment as an example, we can observe the following resource c
 === CPU and RSS memory size
 
 // lint ignore 2 vCPU 16.0 GiB
-Out testing was performed on AWS m7i.large instance type that has 2 vCPUs, 8.0 GiB of Ram and up to 12.5 Gibps of bandwidth. We ingested a 1.1GB single apache JSON log file, running with the throughput and scale preset and self-monitoring enabled.
+We tested using an AWS `m7i.xlarge` instance type with 4 vCPUs, 16.0 GB of memory, and up to 12.5 Gbps of bandwidth. The tests ingested a single log file, running with the throughput preset with self monitoring enabled.
 [options,header]
 |===
 |Resource | Throughput | Scale
@@ -89,7 +89,7 @@ Out testing was performed on AWS m7i.large instance type that has 2 vCPUs, 8.0 G
 | **Read network throughput** | ~40 KB/s | ~2.4 KB/s
 |===
 
-* including all monitoring processes
+\* including all monitoring processes
 
 Adding integrations will increase the memory used by the agent and its processes.
 
--- a/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
@@ -79,7 +79,8 @@ Using our lab environment as an example, we can observe the following resource c
 === CPU and RSS memory size
 
 // lint ignore 2 vCPU 16.0 GiB
-We tested using an AWS `m7i.xlarge` instance type with 4 vCPUs, 16.0 GB of memory, and up to 12.5 Gbps of bandwidth. The tests ingested a single log file using both the throughput and scale preset with self monitoring enabled. These tests are representative of use cases that attempt to ingest data as fast as possible. This does not represent the resource overhead when using {integrations-docs}/endpoint[{elastic-defend}].
+We tested using an AWS `m7i.xlarge` instance type with 4 vCPUs, 16.0 GB of memory, and up to 12.5 Gbps of bandwidth. The tests ingested a single log file using both the <<output-elasticsearch-performance-tuning-settings,throughput and scale preset>> with self monitoring enabled.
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
```diff
@@ -68,21 +68,24 @@ If you are using {agent} with link:{serverless-docs}[{serverless-full}], note th
...
-The following are the minimum system requirements for installing {agent}.
 // lint ignore mem
 [discrete]
 === CPU and RSS memory size
 
-// lint ignore 2vcpu 1gb
-Minimum requirements have been determined by running the {agent} on a GCP `e2-micro` instance (2vCPU/1GB).
-The {agent} used the default policy, running the system integration and self-monitoring.
-
+// lint ignore 4 vCPU 16.0 GiB We tested in AWS m7i.xlarge instance type that has 4 vCPUs, 16.0 GiB of memory and up to 12.5 Gibps of bandwidth. Ingesting a single log file, running with the throughput preset with self-monitoring enabled.
```
kilfoyle, line 81:
```suggestion
// lint ignore 4 vCPU 16.0 GiB 
We tested using an AWS `m7i.xlarge` instance type with 4 vCPUs, 16.0 GB of memory, and up to 12.5 Gbps of bandwidth.
The tests Ingested a single log file, running with the throughput preset with self-monitoring enabled.
```

@alexsapran, Thanks for adding this! I hope you don't mind a comment while the PR is still a draft. Here's how the output will render:

Screenshot 2024-10-31 at 10 05 09 AM

docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
```diff
@@ -12,7 +12,7 @@ run the commands described here. After the {agent} service is installed and runn
 make sure you run these commands without prepending them with `./` to avoid
 invoking the wrong binary.
 * Running {agent} commands using the Windows PowerShell ISE is not supported.
-* See also the <<elastic-agent-installation-minimum-requirements,minimum requirements>> described on this page. 
+* See also the <<elastic-agent-installation-minimum-requirements,minimum requirements>> described on this page.
```
rowlandgeoff, line 15: If we're changing from "minimum" to "resource", we should update the link target and its text too.

rowlandgeoff:
I could suggest some updates to wording and formatting, but first I think we should confirm whether this is the data we'd like current and potential customers to see. Going from 2% to 40% CPU usage is a big jump. Should the default preset info be shown as well to indicate that not every scenario is resource intensive? @cmacknz @amitkanfer

docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
alexsapran: The anchor was not changed, but you are right; it should be renamed, and to rename the reference.

docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
alexsapran: No, not at all. I was hoping to render like that and keep the previous table format. 

If we feel we should adjust to something else, let me know, I'm happy to adjust.

cmacknz:
We effectively have two use cases we care about, and now that I've seen this we probably need to publish results for both of them:
1. High Throughput: How much data can we push assuming we can utilize all the resources of the machine. This is what we have here.
2. Scale/Endpoint Security: What is the resource footprint for low volume security monitoring. This is a competitive metric for security deployments.

We do not want to publish this without a callout that it is exclusive of use case 2.

docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
kilfoyle: Looks good! My only suggestion is to rephrase line 81 a bit:

```
We tested using an AWS `m7i.xlarge` instance type with 4 vCPUs, 16.0 GB of memory, and up to 12.5 Gbps of bandwidth.
The tests ingested a single log file, running with the throughput preset with self monitoring enabled.
```

Apart from that I think it's perfect! 👍

docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
```diff
@@ -57,32 +57,40 @@ Refer to:
...
 // lint ignore mem
 [discrete]
 === CPU and RSS memory size
 
-// lint ignore 2vcpu 1gb
-Minimum requirements have been determined by running the {agent} on a GCP `e2-micro` instance (2vCPU/1GB).
-The {agent} used the default policy, running the system integration and self-monitoring.
-
+// lint ignore 2 vCPU 16.0 GiB
+We tested using an AWS `m7i.xlarge` instance type with 4 vCPUs, 16.0 GB of memory, and up to 12.5 Gbps of bandwidth. The tests ingested a single log file, running with the throughput preset with self monitoring enabled.
```
cmacknz, line 82:
```suggestion
We tested using an AWS `m7i.xlarge` instance type with 4 vCPUs, 16.0 GB of memory, and up to 12.5 Gbps of bandwidth. The tests ingested a single log file using both the throughput and scale preset with self monitoring enabled. These tests are representative of use cases that attempt to ingest data as fast as possible. This does not represent the resource overhead when using [Elastic Defend](https://www.elastic.co/docs/current/integrations/endpoint).
```

docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
```diff
@@ -79,7 +79,7 @@ Using our lab environment as an example, we can observe the following resource c
 === CPU and RSS memory size
 
 // lint ignore 2 vCPU 16.0 GiB
-We tested using an AWS `m7i.xlarge` instance type with 4 vCPUs, 16.0 GB of memory, and up to 12.5 Gbps of bandwidth. The tests ingested a single log file, running with the throughput preset with self monitoring enabled.
+We tested using an AWS `m7i.xlarge` instance type with 4 vCPUs, 16.0 GB of memory, and up to 12.5 Gbps of bandwidth. The tests ingested a single log file using both the throughput and scale preset with self monitoring enabled. These tests are representative of use cases that attempt to ingest data as fast as possible. This does not represent the resource overhead when using [Elastic Defend](https://www.elastic.co/docs/current/integrations/endpoint).
```
kilfoyle, line 82:
```suggestion
We tested using an AWS `m7i.xlarge` instance type with 4 vCPUs, 16.0 GB of memory, and up to 12.5 Gbps of bandwidth. The tests ingested a single log file using both the throughput and scale preset with self monitoring enabled. These tests are representative of use cases that attempt to ingest data as fast as possible. This does not represent the resource overhead when using {integrations-docs}/endpoint[{elastic-defend}].
```

This just converts the link to Asciidoc style and uses our variable for "Elastic Defend".

amitkanfer:
@alexsapran worth adding a link to our presets whenever mentioned...?

strawgate:
> We effectively have two use cases we care about, and now that I've seen this we probably need to publish results for both of them:

Agree, and we should probably present these use-cases separately, as they are quite different.

Also, without something like "events per second" or "disk read throughput", this information doesn't really help with sizing as it stands right now. I'm also a bit worried that we were disk throughput limited in the benchmark, do we have any info on the disk info?

docs/en/ingest-management/elastic-agent/install-elastic-agent.asciidoc
```diff
@@ -57,32 +57,41 @@ Refer to:
...
 // lint ignore mem
 [discrete]
 === CPU and RSS memory size
 
-// lint ignore 2vcpu 1gb
-Minimum requirements have been determined by running the {agent} on a GCP `e2-micro` instance (2vCPU/1GB).
-The {agent} used the default policy, running the system integration and self-monitoring.
-
+// lint ignore 2 vCPU 16.0 GiB
+We tested using an AWS `m7i.xlarge` instance type with 4 vCPUs, 16.0 GB of memory, and up to 12.5 Gbps of bandwidth. The tests ingested a single log file using both the <<output-elasticsearch-performance-tuning-settings,throughput and scale preset>> with self monitoring enabled.
```
alexsapran, line 82: remind to self update the instance type, this instance type is the initial testing

kilfoyle (approved):
@alexsapran This looks great to me so far. I'll be offline until after next week so if you need any help or an approval from `platform-docs` @karenzone can help.

And, I look forward to working with you on automating some of this content eventually. ⚡

amitkanfer (approved):

rowlandgeoff (approved):

Issue #1379, opened 2024-10-09
[Kibana Docs] Broken link on Kibana Fleet APIs documentation


leandrojmp:
Hello,

In the [Kibana Fleet APIs documentation ](https://www.elastic.co/guide/en/fleet/current/fleet-api-docs.html) there is a link to the spec file of the Fleet API, but this link is now broken, it leads to a non existent file on the Kibana repository.

This is the link: [https://github.com/elastic/kibana/blob/main/x-pack/plugins/fleet/common/openapi/bundled.json](https://github.com/elastic/kibana/blob/main/x-pack/plugins/fleet/common/openapi/bundled.json)

It seems that the only file in this path is now the README.md file.

kpollich:
Great catch and thank you for filing this @leandrojmp - I've transferred this to our docs repo. 

cc @kilfoyle - Looks like we broke this link when we swapped around some of the OpenAPI code gen stuff 😶. Would you be able to help with fixing this?

kilfoyle:
@lcawl Now that we have published, generated API docs 🎉 , would you be able to point me to the Fleet API docs and also the spec? I'll update the links in the [Kibana Fleet APIs](https://www.elastic.co/guide/en/fleet/current/fleet-api-docs.html) page.

---

Kibana Fleet APIs
[edit](https://github.com/elastic/ingest-docs/edit/8.15/docs/en/ingest-management/fleet/fleet-api-docs.asciidoc)
You can find details for all available Fleet API endpoints in our generated [Fleet API docs](https://www.elastic.co/guide/en/fleet/current/fleet-apis.html). This documentation is experimental and may be incomplete or change later.

The main source of truth for the Fleet API can be found in the [self-contained spec file](https://github.com/elastic/kibana/blob/main/x-pack/plugins/fleet/common/openapi/bundled.json) that you can use to generate docs using Swagger or a similar tool. For more information, refer to the [Fleet OpenAPI readme](https://github.com/elastic/kibana/tree/main/x-pack/plugins/fleet/common/openapi/README.md).

In this section, we provide examples of some commonly used Fleet APIs.

lcawl:
> [@lcawl](https://github.com/lcawl) Now that we have published, generated API docs 🎉 , would you be able to point me to the Fleet API docs and also the spec?

Since there are multiple groupings of Fleet APIs, it's simplest to just point to "...the generated {api-kibana}[Kibana API docs]".

[...]

PR #1415, opened 2024-10-31
Fix API docs link

kilfoyle:
This updates the first paragraph of the [Kibana Fleet APIs](https://www.elastic.co/guide/en/fleet/current/fleet-api-docs.html) page to fix the broken link and point to the new generated API docs [here](https://www.elastic.co/docs/api/doc/kibana).

Screenshot 2024-10-31 at 12 55 22 PM

Closes: https://github.com/elastic/ingest-docs/issues/1379

Note: I've left the examples on the page in place, but if these should be revised or moved, I'm happy to do that (separately from this PR).


juliaElastic (approved):

karenzone (approved):
Nice! Thank you.
LGTM

PR #1430, opened 2024-11-04
Move section about Logstash to bottom of Kafka output page

kilfoyle:
At the top of the [Kafka output settings](https://www.elastic.co/guide/en/fleet/current/kafka-output-settings.html#_kafka_output_and_using_logstash_to_index_data_to_elasticsearch) page is a section warning users about potential problems when they use use Logstash to modify the Kafka output on its way to Elasticsearch. It's not ideal for all users to start their Kafka output journey with that section.

I think it'd be better to have that section at the bottom of the Kafka output settings page, and to make sure people find the section we can add a warning at the top of the page, like this:

screen

---

The guidance for Logstash appears just as before, but at the bottom of the page:

---

Screenshot 2024-11-04 at 5 53 08 PM

---

@lucabelluccini, @nimarezainia  Any thoughts or concerns?


karenzone (approved):
LGTM

lucabelluccini (approved):

PR #1465, opened 2024-11-14
Update bulk_reassign_agents_request to include batchSize body parameter

asmith-elastic:
This change documents the ability to leverage a `batchSize` body parameter for the `bulk_reassign_agents_request` which was introduced in the following PR: https://github.com/elastic/kibana/pull/134565

This will also help align the documentation with the following Kibana API docs: https://www.elastic.co/docs/api/doc/kibana/operation/operation-post-fleet-agents-bulk-reassign


kilfoyle (approved):
Thanks @asmith-elastic! I'll backport this into the 8.16 docs.

The page being updated is generated, but I guess we won't be generating them anymore, and likely removing them in favour of the newer, better ones that you pointed to in the description. :-)

Issue #1503, opened 2024-11-27
Add docs for pre-existing Active Directory user for unprivileged agent


kilfoyle:
Mark as technical preview until AD support can be fully tested.

Use case:

>For example, if I have a service account in active dieectory that works on the system already, can I use that or is elastic creating an account and group locally?
>
>The use case is integrations that need network access from specific service accounts in active directory to be the service context for filebeat (api, file shares, etc).
>
>Today system is the only option but this gives me hope for 1 non system which is great, but having custom users would be excellent.
>
>In the past you could simply install filebeat or winglogbeat with the service account you needed as a service and it would use that user context when it runs.

Details are [here](https://github.com/elastic/elastic-agent/pull/5988#issuecomment-2502464911) and [here](https://github.com/elastic/elastic-agent/issues/4585#issuecomment-2502387685).

PR #1505, opened 2024-11-27
Add docs for running unprivileged agent as existing user/group

kilfoyle:
This adds docs for running Elastic Agent unprivileged mode, as a specific user or group (e.g. under an Active Directory account).

Targets: 8.x
Closes: #1503

---

**Update to [Run Elastic Agent without administrative privileges](https://www.elastic.co/guide/en/fleet/master/elastic-agent-unprivileged.html) page:**

screen1

---

**Update to Command Reference [`elastic-agent install` command](https://www.elastic.co/guide/en/fleet/current/elastic-agent-cmd-options.html#elastic-agent-install-command) `--unprivileged` option:**

screen3

---

**Update to Command Reference [`elastic-agent unprivileged` command](https://www.elastic.co/guide/en/fleet/current/elastic-agent-cmd-options.html#elastic-agent-unprivileged-command):**

screen2

### DIFF
--- a/docs/en/ingest-management/commands.asciidoc
+++ b/docs/en/ingest-management/commands.asciidoc
@@ -1191,7 +1191,7 @@ For details and limitations for running {agent} in this mode, refer to {fleet-gu
 
 Note that changing a running {agent} to `unprivileged` mode is prevented if the agent is currently enrolled with a policy that contains the {elastic-defend} integration.
 
-[technical preview] To run {agent} without superuser privileges as a pre-existing user or group, for instance under an Active Directory account, add the `--user` and `--password` parameters.
+[technical preview] To run {agent} without superuser privileges as a pre-existing user or group, for instance under an Active Directory account, add either a `--user` or `--group` parameter together with a `--password` parameter.
 
 [discrete]
 === Examples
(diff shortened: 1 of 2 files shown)


nimarezainia (approved):

michalpristas (approved):
this looks good, many thanks @kilfoyle

docs/en/ingest-management/commands.asciidoc
```diff
@@ -823,8 +823,22 @@ tags, you must unenroll the {agent}, then re-enroll it using new tags.
 Run {agent} without full superuser privileges.
 This option is useful in organizations that limit `root` access on Linux or macOS systems, or `admin` access on Windows systems.
 For details and limitations for running {agent} in this mode, refer to {fleet-guide}/elastic-agent-unprivileged.html[Run {agent} without administrative privileges].
-
++
 Note that changing to `unprivileged` mode is prevented if the agent is currently enrolled in a policy that includes an integration that requires administrative access, such as the {elastic-defend} integration.
++
+[technical preview] To run {agent} without superuser privileges as a pre-existing user or group, for instance under an Active Directory account, you can specify the user or group, and the password to use.
```
lcawl, line 829: > [technical preview] To run {agent} without superuser privileges as a pre-existing user or group, for instance under 

This makes me curious why/if https://github.com/elastic/docs?tab=readme-ov-file#using-the-technical-preview-admonition doesn't work in this situation.

lcawl (approved):
One comment of curiosity, otherwise LGTM

docs/en/ingest-management/commands.asciidoc
kilfoyle: Good catch. The admonitions work, but the big banner doesn't look right in the context of individual command options. So instead I went with the approach we used in the Elasticsearch docs for a similar problem.

Issue #1511, opened 2024-12-02
Improve titles in "Deployment models" section


kilfoyle:
Feeback from Edu (via Slack) about the current Fleet Docs Information Architecture:

 - It's a bit difficult to find where the Fleet Server installation guides are located, they are under "Deployment models", and in the past they were under "Set up Fleet Server" section.
 - I like the deployment models section, but what do you think of renaming it a bit to `Fleet Deployment Models` or something like that?
 - Also in the sub-sections the third one is missing `Fleet Server`, and the forth one has it. I think they should be aligned and keep names consistency (my vote is to include fleet server always, otherwise the users won't really know that it's the doc to install fleet server).

![Image](https://github.com/user-attachments/assets/c67497e1-023d-4990-b7b7-ae3f043736f1)

kilfoyle:
Closing. Based on discussion in the [PR](https://github.com/elastic/ingest-docs/pull/1512) we've decided to leave the titles as is.

PR #1512, opened 2024-12-02
Update titles in 'Deployment models' section

kilfoyle:
Edu had [some feedback](https://github.com/elastic/ingest-docs/issues/1511) on improving the titles in the "Deployment models" section to make more clear that this is about Fleet, and also to improve consistency.

Closes: https://github.com/elastic/ingest-docs/issues/1511

---

**Original:**
Screenshot 2024-12-02 at 10 13 47 AM

**Updated:**
Screenshot 2024-12-02 at 10 12 01 AM

### DIFF
--- a/docs/en/ingest-management/fleet/add-fleet-server-cloud.asciidoc
+++ b/docs/en/ingest-management/fleet/add-fleet-server-cloud.asciidoc
@@ -1,5 +1,5 @@
 [[add-fleet-server-cloud]]
-= Deploy on {ecloud}
+= Deploy {fleet-server} on {ecloud}
 
 To use {fleet} for central management, a <<fleet-server,{fleet-server}>> must
 be running and accessible to your hosts.
--- a/docs/en/ingest-management/fleet/add-fleet-server-on-prem.asciidoc
+++ b/docs/en/ingest-management/fleet/add-fleet-server-on-prem.asciidoc
@@ -1,5 +1,5 @@
 [[add-fleet-server-on-prem]]
-= Deploy on-premises and self-managed
+= Deploy {fleet-server} on-premises and self-managed
 
 To use {fleet} for central management, a <<fleet-server,{fleet-server}>> must
 be running and accessible to your hosts.
--- a/docs/en/ingest-management/fleet/fleet-deployment-models.asciidoc
+++ b/docs/en/ingest-management/fleet/fleet-deployment-models.asciidoc
@@ -1,5 +1,5 @@
 [[fleet-deployment-models]]
-= Deployment models
+= {fleet} Deployment models
 
 There are various models for setting up {agents} to work with {es}.
 The recommended approach is to use {fleet}, a web-based UI in Kibana, to centrally manage all of your {agents} and their policies.


georgewallace (approved):

eedugon (approved):
LGTM

kilfoyle:
@eedugon  I spoke with @nimarezainia about this and we're thinking that the change I have above might not be good:

 - "Fleet deployment models" - The deployment isn't specifically about Fleet, since Fleet is only the management UI. So perhaps that title should remain as is, i.e, "Deployment models".
 - "Deploy Fleet Server on Elastic Cloud" and "Deploy Fleet Server on premises and self-managed" - Similarly, these pages aren't only about Fleet Server.

I think you're right that the titles are inconsistent. So instead maybe we should change "Deploy Fleet Server on premises and  Elasticsearch on Cloud" to just "Deploy on premises and Elasticsearch on Cloud", like this:
  Screenshot 2024-12-12 at 11 21 04 AM


What do you think?

eedugon:
> "Fleet deployment models" - The deployment isn't specifically about Fleet, since Fleet is only the management UI. So perhaps that title should remain as is, i.e, "Deployment models".

I totally agree on that.
Actually everything is about Fleet (not UI), as the entire book is the `fleet` book. So `Deployment models` or `Fleet deployment models` are totally ok to me.

> "Deploy Fleet Server on Elastic Cloud" and "Deploy Fleet Server on premises and self-managed" - Similarly, these pages aren't only about Fleet Server.

mmm I clearly disagree here :)

IMHO, those 2 docs are 95% / 98% about Fleet Server only, and Fleet Server installation is the clear and main goal of the docs. On those two docs, **everything**, all the paragraphs and sections, are about the Fleet Server. Except very minor sentences or lines.

Of course to install the fleet server we have to setup fleet (a policy, an integration, etc), but that's like part of the process that has the goal of installing / deploying the fleet server itself. Also those docs don't cover the installation of the future Elastic Agents, so more reasons to consider them focused mainly on the Fleet Server component of the Fleet ecosystem.

Anyway, I don't think these names are super important, and I'm happy with whatever you and @nimarezainia prefer, but it's just that the current names have created confusion, and I'll share an example:

Colleagues from support asked me a few weeks ago where we had the Fleet Server installation docs (for on-premises), as they were clearly identified in 7.17 docs and they couldn't find them in the current docs.
After searching and navigating through the docs I actually thought we didn't have those docs anymore :D... until I realized that we actually had 2 docs for Fleet Server installation:
- Deploy on premises
- Deploy on premises with Elasticsearch on cloud

None of the names helped anyone to realize that they were about Deploying / Installing the Fleet Server, it's not intuitive at all when the docs are 95% about that.

But as I said I'm ok with cancelling the change.

the question now is... I'm creating the doc `Deploying Fleet Server on Kubernetes`.... should we call it `Deploy Fleet on Kubernetes`? It will be the same, we can call it Deploy on Kubernetes, but it's a super ambiguous name where it's almost everything about deploying a Fleet Server on Kubernetes.

kilfoyle:
@nimarezainia just FYI in case you have thoughts on the "deployment models" naming and Edu's comment above.

eedugon:
@nimarezainia , @kilfoyle : I agree with @nimarezainia 's view, and I think it's ok if we don't change anything.

The original `inconsistency` is maybe not an inconsistency:

image

The third one specifies `Fleet Server` because one of the stack components will be on premises while Elasticsearch will be on the cloud, so it might have sense to explicitly mention it. As also @nimarezainia mentioned it's the `distinguishing factor`.

And about the lack of `fleet server` in the other titles I think it's all a matter of taste, nothing serious.

So, whatever you prefer, we can:
- Cancel this PR
- Apply the change proposed by @kilfoyle by removing the `fleet server statement`.

As soon as we agree I will update (or cancel) this PR. Thanks!

nimarezainia:
thanks @eedugon - I'm ok cancelling this and concentrating efforts elsewhere. This section's intention was to discuss different deployment models. Fleet server is a component that generally changes where it's deployed in each model - so mentioning that in the titles is fine.

kilfoyle:
Thanks a lot for the input Edu and Nima! Given the explanations it makes sense to me to leave things as is, so I'll cancel this one.

PR #1518, opened 2024-12-03
Fleet Server on kubernetes document proposal

eedugon:
Adding a `Deploy Fleet Server on Kubernetes` document, under `Deployment models` section.

Preview: https://ingest-docs_bk_1518.docs-preview.app.elstc.co/guide/en/fleet/master/add-fleet-server-kubernetes.html

### DIFF
--- a/docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
+++ b/docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
@@ -153,7 +153,7 @@ spec:
             value: /mnt/certs/es-ca.crt
           - name: FLEET_SERVER_CERT
             value: /mnt/certs/fleet-server.crt
-          - name: FLEET_SERVER_KEY
+          - name: FLEET_SERVER_CERT_KEY
             value: /mnt/certs/fleet-server.key
         ports:
         - containerPort: 8220
--- a/docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
+++ b/docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
@@ -111,7 +111,7 @@ spec:
 // ***************************************************
 
 // tag::production-deployment[]
-[source, yaml]
+["source","yaml",subs="attributes"]
 ------------------------------------------------------------
 apiVersion: apps/v1
 kind: Deployment
--- a/docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
+++ b/docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
@@ -1,4 +1,6 @@
 // tag::quickstart-secret[]
+The following command assumes you have the {es} CA file available in a local file.
++
 [source, shell]
 ------------------------------------------------------------
 kubectl create secret generic fleet-server-certs \
@@ -13,6 +15,8 @@ If you prefer to obtain a `yaml manifest` of the Secret to create, append `--dry
 // ***************************************************
 
 // tag::production-secret[]
+The following command assumes you have the {es} CA file, the {fleet-server} certificate, and the {fleet-server} key available in local files.
++
 [source, shell]
 ------------------------------------------------------------
 kubectl create secret generic fleet-server-certs \
--- a/docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
+++ b/docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
@@ -42,7 +42,7 @@ kind: Deployment
 metadata:
   name: fleet-server
 spec:
-  replicas: 1
+  replicas: 1 <1>
   selector:
     matchLabels:
       app: fleet-server
@@ -51,12 +51,12 @@ spec:
       labels:
         app: fleet-server
     spec:
-      automountServiceAccountToken: false <1>
+      automountServiceAccountToken: false <2>
       containers:
       - name: elastic-agent
         image: docker.elastic.co/beats/elastic-agent:{version}
         env:
-          - name: FLEET_SERVER_ENABLE <2>
+          - name: FLEET_SERVER_ENABLE
             value: "true"
           - name: FLEET_SERVER_ELASTICSEARCH_HOST
             valueFrom:
@@ -90,9 +90,9 @@ spec:
           optional: false
           secretName: fleet-server-certs
 ------------------------------------------------------------
-<1> Keep `automountServiceAccountToken` set to `false` to disable the <<kubernetes-provider>>
-<2> Keep `FLEET_SERVER_ENABLE` set to `true`
-<3> Consider configuring requests and limits as a best practice
+<1> You can add replicas to scale the {fleet-server} horizontally.
+<2> Keep `automountServiceAccountToken` set to `false` to disable the <<kubernetes-provider>>.
+<3> Consider configuring requests and limits always as a best practice. Refer to <<fleet-server-scalability>> for more information about resources utilization of the {fleet-server}.
 // +
 // Manifest highlights:
 // +
@@ -118,7 +118,7 @@ kind: Deployment
 metadata:
   name: fleet-server
 spec:
-  replicas: 1
+  replicas: 1 <1>
   selector:
     matchLabels:
       app: fleet-server
@@ -127,12 +127,12 @@ spec:
       labels:
         app: fleet-server
     spec:
-      automountServiceAccountToken: false <1>
+      automountServiceAccountToken: false <2>
       containers:
       - name: elastic-agent
         image: docker.elastic.co/beats/elastic-agent:{version}
         env:
-          - name: FLEET_SERVER_ENABLE <2>
+          - name: FLEET_SERVER_ENABLE
             value: "true"
           - name: FLEET_SERVER_ELASTICSEARCH_HOST
             valueFrom:
@@ -170,7 +170,7 @@ spec:
           optional: false
           secretName: fleet-server-certs
 ------------------------------------------------------------
-<1> Keep `automountServiceAccountToken` set to `false` to disable the <<kubernetes-provider>>
-<2> Keep `FLEET_SERVER_ENABLE` set to `true`
-<3> Consider configuring requests and limits as a best practice
+<1> You can add replicas to scale the {fleet-server} horizontally.
+<2> Keep `automountServiceAccountToken` set to `false` to disable the <<kubernetes-provider>>.
+<3> Consider configuring requests and limits always as a best practice. Refer to <<fleet-server-scalability>> for more information about resources utilization of the {fleet-server}.
 // end::production-deployment[]
--- a/docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
+++ b/docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
@@ -1,0 +1,172 @@
+// tag::quickstart-secret[]
+[source, shell]
+------------------------------------------------------------
+kubectl create secret generic fleet-server-certs \
+  --from-file=es-ca.crt=<PATH_TO_ES_CA_CERT_FILE> <1>
+------------------------------------------------------------
+<1> Substitute `<PATH_TO_ES_CA_CERT_FILE>` with your local file containing the {es} CA(s).
++
+If you prefer to obtain a `yaml manifest` of the Secret to create, append `--dry-run=client -o=yaml` to the command and save the output to a file.
+// end::quickstart-secret[]
+
+// ***************************************************
+// ***************************************************
+
+// tag::production-secret[]
+[source, shell]
+------------------------------------------------------------
+kubectl create secret generic fleet-server-certs \
+  --from-file=es-ca.crt=<PATH_TO_ES_CA_CERT_FILE> \ <1>
+  --from-file=fleet-server.crt=<PATH_TO_FLEET_SERVER_CERT> \ <2>
+  --from-file=fleet-server.key=<PATH_TO_FLEET_SERVER_CERT_KEY> <3>
+------------------------------------------------------------
+<1> Substitute `<PATH_TO_ES_CA_CERT_FILE>` with your local file containing the {es} CA(s).
+<2> Substitute `<PATH_TO_FLEET_SERVER_CERT>` with your local file containing the server TLS certificate for the {fleet-server}.
+<3> Substitute `PATH_TO_FLEET_SERVER_CERT_KEY` with your local file containing the server TLS key for the {fleet-server}.
++
+If you prefer to obtain a `yaml manifest` of the Secret to create, append `--dry-run=client -o=yaml` to the command and save the output to a file.
+// end::production-secret[]
+
+// ***************************************************
+// ***************************************************
+
+// tag::quickstart-deployment[]
+[source, yaml]
+------------------------------------------------------------
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: fleet-server
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: fleet-server
+  template:
+    metadata:
+      labels:
+        app: fleet-server
+    spec:
+      automountServiceAccountToken: false <1>
+      containers:
+      - name: elastic-agent
+        image: docker.elastic.co/beats/elastic-agent:{version}
+        env:
+          - name: FLEET_SERVER_ENABLE <2>
+            value: "true"
+          - name: FLEET_SERVER_ELASTICSEARCH_HOST
+            valueFrom:
+              secretKeyRef:
+                name: fleet-server-config
+                key: elastic_endpoint
+          - name: FLEET_SERVER_SERVICE_TOKEN
+            valueFrom:
+              secretKeyRef:
+                name: fleet-server-config
+                key: elastic_service_token
+          - name: FLEET_SERVER_POLICY_ID
+            valueFrom:
+              secretKeyRef:
+                name: fleet-server-config
+                key: fleet_policy_id
+          - name: ELASTICSEARCH_CA
+            value: /mnt/certs/es-ca.crt
+        ports:
+        - containerPort: 8220
+          protocol: TCP
+        resources: {} <3>
+        volumeMounts:
+        - name: certs
+          mountPath: /mnt/certs
+          readOnly: true
+      volumes:
+      - name: certs
+        secret:
+          defaultMode: 420
+          optional: false
+          secretName: fleet-server-certs
+------------------------------------------------------------
+<1> Keep `automountServiceAccountToken` set to `false` to disable the <<kubernetes-provider>>
+<2> Keep `FLEET_SERVER_ENABLE` set to `true`
+<3> Consider configuring requests and limits as a best practice
+// +
+// Manifest highlights:
+// +
+// * `automountServiceAccountToken: false`: used to disable the Kubernetes provider of the {fleet-server}.
+// * Feel free to adapt the `name` of the Deployment or the Pod template and selector `labels`.
+// * `resources`: Adapt them to your needs.
+// +
+// [NOTE]
+// ====
+// You can use `FLEET_SERVER_POLICY_ID` instead of `FLEET_SERVER_POLICY_NAME`, and `FLEET_SERVER_ELASTICSEARCH_CA_TRUSTED_FINGERPRINT` instead of `ELASTICSEARCH_CA` if preferred.
+// If you opt for `FLEET_SERVER_POLICY_ID`, provide the ID of the policy instead of the name, and if you opt for `FLEET_SERVER_ELASTICSEARCH_CA_TRUSTED_FINGERPRINT` provide the SHA-256 fingerprint of the {es} CA certificate instead of the entire certificate in PEM format.
+// ====
+// end::quickstart-deployment[]
+
+// ***************************************************
+// ***************************************************
+
+// tag::production-deployment[]
+[source, yaml]
+------------------------------------------------------------
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: fleet-server
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: fleet-server
+  template:
+    metadata:
+      labels:
+        app: fleet-server
+    spec:
+      automountServiceAccountToken: false <1>
+      containers:
+      - name: elastic-agent
+        image: docker.elastic.co/beats/elastic-agent:{version}
+        env:
+          - name: FLEET_SERVER_ENABLE <2>
+            value: "true"
+          - name: FLEET_SERVER_ELASTICSEARCH_HOST
+            valueFrom:
+              secretKeyRef:
+                name: fleet-server-config
+                key: elastic_endpoint
+          - name: FLEET_SERVER_SERVICE_TOKEN
+            valueFrom:
+              secretKeyRef:
+                name: fleet-server-config
+                key: elastic_service_token
+          - name: FLEET_SERVER_POLICY_ID
+            valueFrom:
+              secretKeyRef:
+                name: fleet-server-config
+                key: fleet_policy_id
+          - name: ELASTICSEARCH_CA
+            value: /mnt/certs/es-ca.crt
+          - name: FLEET_SERVER_CERT
+            value: /mnt/certs/fleet-server.crt
+          - name: FLEET_SERVER_KEY
+            value: /mnt/certs/fleet-server.key
+        ports:
+        - containerPort: 8220
+          protocol: TCP
+        resources: {} <3>
+        volumeMounts:
+        - name: certs
+          mountPath: /mnt/certs
+          readOnly: true
+      volumes:
+      - name: certs
+        secret:
+          defaultMode: 420
+          optional: false
+          secretName: fleet-server-certs
+------------------------------------------------------------
+<1> Keep `automountServiceAccountToken` set to `false` to disable the <<kubernetes-provider>>
+<2> Keep `FLEET_SERVER_ENABLE` set to `true`
+<3> Consider configuring requests and limits as a best practice
+// end::production-deployment[]
--- a/docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
+++ b/docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
@@ -155,6 +155,10 @@ spec:
             value: /mnt/certs/fleet-server.crt
           - name: FLEET_SERVER_CERT_KEY
             value: /mnt/certs/fleet-server.key
+          - name: FLEET_URL
+            value: <url>
+          - name: FLEET_SERVER_PORT
+            value: '8220'
         ports:
         - containerPort: 8220
           protocol: TCP
--- a/docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
+++ b/docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
@@ -1,34 +1,48 @@
 // tag::quickstart-secret[]
-The following command assumes you have the {es} CA file available in a local file.
+The following command assumes you have the {es} CA available as a local file.
 +
 [source, shell]
 ------------------------------------------------------------
-kubectl create secret generic fleet-server-certs \
-  --from-file=es-ca.crt=<PATH_TO_ES_CA_CERT_FILE> <1>
+kubectl create secret generic fleet-server-ssl \
+  --from-file=es-ca.crt=<PATH_TO_ES_CA_CERT_FILE>
 ------------------------------------------------------------
-<1> Substitute `<PATH_TO_ES_CA_CERT_FILE>` with your local file containing the {es} CA(s).
 +
-If you prefer to obtain a `yaml manifest` of the Secret to create, append `--dry-run=client -o=yaml` to the command and save the output to a file.
+--
+When running the command, substitute the following values:
+
+* `<PATH_TO_ES_CA_CERT_FILE>` with your local file containing the {es} CA(s).
+--
++
+If you prefer to obtain a *yaml manifest* of the Secret to create, append `--dry-run=client -o=yaml` to the command and save the output to a file.
 // end::quickstart-secret[]
 
 // ***************************************************
 // ***************************************************
 
 // tag::production-secret[]
-The following command assumes you have the {es} CA file, the {fleet-server} certificate, and the {fleet-server} key available in local files.
+The following command assumes you have the {es} CA and the {fleet-server} certificate, key and CA available as local files.
 +
 [source, shell]
 ------------------------------------------------------------
-kubectl create secret generic fleet-server-certs \
-  --from-file=es-ca.crt=<PATH_TO_ES_CA_CERT_FILE> \ <1>
-  --from-file=fleet-server.crt=<PATH_TO_FLEET_SERVER_CERT> \ <2>
-  --from-file=fleet-server.key=<PATH_TO_FLEET_SERVER_CERT_KEY> <3>
+kubectl create secret generic fleet-server-ssl \
+  --from-file=es-ca.crt=<PATH_TO_ES_CA_CERT_FILE> \
+  --from-file=fleet-ca.crt=<PATH_TO_FLEET_CA_CERT_FILE> \
+  --from-file=fleet-server.crt=<PATH_TO_FLEET_SERVER_CERT> \
+  --from-file=fleet-server.key=<PATH_TO_FLEET_SERVER_CERT_KEY> \
+  --from-literal=fleet_url='<FLEET_URL>'
 ------------------------------------------------------------
-<1> Substitute `<PATH_TO_ES_CA_CERT_FILE>` with your local file containing the {es} CA(s).
-<2> Substitute `<PATH_TO_FLEET_SERVER_CERT>` with your local file containing the server TLS certificate for the {fleet-server}.
-<3> Substitute `PATH_TO_FLEET_SERVER_CERT_KEY` with your local file containing the server TLS key for the {fleet-server}.
 +
-If you prefer to obtain a `yaml manifest` of the Secret to create, append `--dry-run=client -o=yaml` to the command and save the output to a file.
+--
+When running the command, substitute the following values:
+
+* `<PATH_TO_ES_CA_CERT_FILE>` with your local file containing the {es} CA(s).
+* `<PATH_TO_FLEET_CA_CERT_FILE>` with your local file containing the {fleet-server} CA.
+* `<PATH_TO_FLEET_SERVER_CERT>` with your local file containing the server TLS certificate for the {fleet-server}.
+* `<PATH_TO_FLEET_SERVER_CERT_KEY>` with your local file containing the server TLS key for the {fleet-server}.
+* `<FLEET_URL>` with the URL that points to the {fleet-server}, for example `https://fleet-svc`. This URL will be used by the {fleet-server} during its bootstrap, and its hostname must be included in the server certificate’s x509 Subject Alternative Name (SAN) list.
+--
++
+If you prefer to obtain a *yaml manifest* of the Secret to create, append `--dry-run=client -o=yaml` to the command and save the output to a file.
 // end::production-secret[]
 
 // ***************************************************
@@ -37,12 +51,25 @@ If you prefer to obtain a `yaml manifest` of the Secret to create, append `--dry
 // tag::quickstart-deployment[]
 ["source","yaml",subs="attributes"]
 ------------------------------------------------------------
+apiVersion: v1
+kind: Service
+metadata:
+  name: fleet-svc
+spec:
+  type: ClusterIP
+  selector:
+    app: fleet-server
+  ports:
+  - port: 443
+    protocol: TCP
+    targetPort: 8220
+---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: fleet-server
 spec:
-  replicas: 1 <1>
+  replicas: 1
   selector:
     matchLabels:
       app: fleet-server
@@ -51,7 +78,7 @@ spec:
       labels:
         app: fleet-server
     spec:
-      automountServiceAccountToken: false <2>
+      automountServiceAccountToken: false
       containers:
       - name: elastic-agent
         image: docker.elastic.co/beats/elastic-agent:{version}
@@ -78,7 +105,7 @@ spec:
         ports:
         - containerPort: 8220
           protocol: TCP
-        resources: {} <3>
+        resources: {}
         volumeMounts:
         - name: certs
           mountPath: /mnt/certs
@@ -88,23 +115,8 @@ spec:
         secret:
           defaultMode: 420
           optional: false
-          secretName: fleet-server-certs
+          secretName: fleet-server-ssl
 ------------------------------------------------------------
-<1> You can add replicas to scale the {fleet-server} horizontally.
-<2> Keep `automountServiceAccountToken` set to `false` to disable the <<kubernetes-provider>>.
-<3> Consider configuring requests and limits always as a best practice. Refer to <<fleet-server-scalability>> for more information about resources utilization of the {fleet-server}.
-// +
-// Manifest highlights:
-// +
-// * `automountServiceAccountToken: false`: used to disable the Kubernetes provider of the {fleet-server}.
-// * Feel free to adapt the `name` of the Deployment or the Pod template and selector `labels`.
-// * `resources`: Adapt them to your needs.
-// +
-// [NOTE]
-// ====
-// You can use `FLEET_SERVER_POLICY_ID` instead of `FLEET_SERVER_POLICY_NAME`, and `FLEET_SERVER_ELASTICSEARCH_CA_TRUSTED_FINGERPRINT` instead of `ELASTICSEARCH_CA` if preferred.
-// If you opt for `FLEET_SERVER_POLICY_ID`, provide the ID of the policy instead of the name, and if you opt for `FLEET_SERVER_ELASTICSEARCH_CA_TRUSTED_FINGERPRINT` provide the SHA-256 fingerprint of the {es} CA certificate instead of the entire certificate in PEM format.
-// ====
 // end::quickstart-deployment[]
 
 // ***************************************************
@@ -113,12 +125,25 @@ spec:
 // tag::production-deployment[]
 ["source","yaml",subs="attributes"]
 ------------------------------------------------------------
+apiVersion: v1
+kind: Service
+metadata:
+  name: fleet-svc
+spec:
+  type: ClusterIP
+  selector:
+    app: fleet-server
+  ports:
+  - port: 443
+    protocol: TCP
+    targetPort: 8220
+---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: fleet-server
 spec:
-  replicas: 1 <1>
+  replicas: 1
   selector:
     matchLabels:
       app: fleet-server
@@ -127,7 +152,7 @@ spec:
       labels:
         app: fleet-server
     spec:
-      automountServiceAccountToken: false <2>
+      automountServiceAccountToken: false
       containers:
       - name: elastic-agent
         image: docker.elastic.co/beats/elastic-agent:{version}
@@ -155,14 +180,21 @@ spec:
             value: /mnt/certs/fleet-server.crt
           - name: FLEET_SERVER_CERT_KEY
             value: /mnt/certs/fleet-server.key
+          - name: FLEET_CA
+            value: /mnt/certs/fleet-ca.crt
           - name: FLEET_URL
-            value: <url>
+            valueFrom:
+              secretKeyRef:
+                name: fleet-server-ssl
+                key: fleet_url
+          - name: FLEET_SERVER_TIMEOUT
+            value: '60s'
           - name: FLEET_SERVER_PORT
             value: '8220'
         ports:
         - containerPort: 8220
           protocol: TCP
-        resources: {} <3>
+        resources: {}
         volumeMounts:
         - name: certs
           mountPath: /mnt/certs
@@ -172,9 +204,6 @@ spec:
         secret:
           defaultMode: 420
           optional: false
-          secretName: fleet-server-certs
+          secretName: fleet-server-ssl
 ------------------------------------------------------------
-<1> You can add replicas to scale the {fleet-server} horizontally.
-<2> Keep `automountServiceAccountToken` set to `false` to disable the <<kubernetes-provider>>.
-<3> Consider configuring requests and limits always as a best practice. Refer to <<fleet-server-scalability>> for more information about resources utilization of the {fleet-server}.
 // end::production-deployment[]
--- a/docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
+++ b/docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
@@ -335,7 +335,7 @@ kubectl logs fleet-server-xxxxx
 +
 It could take a while until the {fleet-server} fully starts. If you left the {kib} browser window opened during <<add-fleet-server-kubernetes-preparations>> it will show *Connected* in case everything goes well.
 +
-Do not continue with the next steps if the {fleet-server} is not completely up and healthy and appears in Fleet UI.
+Do not continue with the next steps if the {fleet-server} is not completely up and healthy, and appearing in {fleet} UI.
 
 . Expose the {fleet-server} to {agent}s through the *Fleet Server Host*:
 +
--- a/docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
+++ b/docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
@@ -24,57 +24,47 @@ To deploy a self-managed {fleet-server}, you need to:
 bugfix releases)
 ** {kib} should be on the same minor version as {es}.
 
-* {ece} 2.9 or later
-+
---
-** Requires additional wildcard domains and certificates (which normally only
-cover `*.cname`, not `*.*.cname`). This enables us to provide the URL for
-{fleet-server} of `https://.fleet.`.
-** The deployment template must contain an {integrations-server} node.
---
-+
-For more information about hosting {fleet-server} on {ece}, refer to
-{ece-ref}/ece-manage-integrations-server.html[Manage your {integrations-server}].
-
 [discrete]
 [[add-fleet-server-kubernetes-prereq]]
 = Prerequisites
 
 Before deploying, you need to:
 
-* Plan the SSL setup, server certificates, {fleet-server} URL, and CAs.
+* Plan the SSL setup, server certificates, {fleet-server} URL, and certificate authorities (CAs).
 * Ensure components have access to the ports needed for communication.
 
 [discrete]
 [[add-fleet-server-kubernetes-cert-prereq]]
 == SSL certificates considerations
 
-This section shows the minimum requirements in terms of Transport Layer Security (TLS) certificates for the {fleet-server}, assuming no `mutual TLS` (mTLS) is needed. Refer to https://LINK[mTLS setup] if needed. The traffic flows described in this section support mTLS.
+This section shows the minimum requirements in terms of Transport Layer Security (TLS) certificates for the {fleet-server}, assuming no mutual TLS (mTLS) is needed. Refer to <<tls-overview>> and <<mutual-tls>> for more information about the configuration needs of both approaches.
 
 There are two main and differentiated traffic flows affecting the {fleet-server}:
 
 * *{agent} -> {fleet-server}* traffic flow:
 +
-In this flow the {fleet-server} acts as the server and the {agent} acts as the client, therefore the {fleet-server} requires an SSL certificate and key, and the {agent} needs to trust the CA certificate used to sign the certificate.
+In this flow the {fleet-server} acts as the server and the {agent} acts as the client, therefore the {fleet-server} requires an SSL certificate and key, and the {agent} needs to trust the CA certificate used to sign the {fleet-server} certificate.
 +
 If your organization already uses the {stack}, you may already have a CA certificate that could be used to generate the new cert for the {fleet-server}. If you do not have a CA certificate, you can read more
-about generating one in <<secure-connections>>.
+about generating a CA and a server certificate for the {fleet-server} in <<generate-fleet-server-certs>>.
 +
-Ensure you also add any of the {fleet-server} `FQDN`, `hostname`, or `ip address` that will be used by {agent}s to access the {fleet-server} in the server certificate as an `x.509 Subject Alternative Name (SAN)`. On Kubernetes this will depend on the method used to expose the {fleet-server} and your specific environment.
+You also need to know and plan in advance the `URL` (or `URLS`) that the {agent} clients will use to access the {fleet-server}, because the *hostname* part of the URL needs to be included in the server certificate as an `x.509 Subject Alternative Name (SAN)`. If you plan to make your {fleet-server} accessible in multiple *hostnames*, add all of them to the server certificate.
 +
-NOTE: The {fleet-server} certificate is not required when testing and iterating using the *Quick start* option, but should always be used for *production* deployments.
+NOTE: The {fleet-server} certificate is not required when installing the server using the *Quick start* mode, but should always be used for *production* deployments. In *Quick start* mode, the {fleet-server} uses a self-signed certificate and the {agent}s have to be enrolled with the `--insecure` option.
 
 * *{fleet-server} -> {es} output* traffic flow: 
 +
 The {fleet-server} acts as the client and {es} acts as the HTTPS server. The {fleet-server} needs to trust the CA certificate used to sign the {es} certificate. If your {es} is using certificates signed by a corporate CA or by multiple intermediate CAs you will need to use them during the {fleet-server} setup.
 +
 NOTE: If your {es} cluster is on Elastic Cloud or if it uses a certificate signed by a public and known CA you won't need the {es} CA during the setup.
++
+TBD: should we mention that the CA has to be in PEM format or provide a `SHA-256 fingerprint` of the CA?
 
 In summary, you need:
 
-* A `server certificate` and `key`, configured for the `{fleet-server} URL`, used to encrypt traffic between the {fleet-server}s and the {agent}s. The CA used to sign this certificate will be needed by the {agent}s, not by the {fleet-server} itself.
+* A *server certificate and key*, valid for the {fleet-server} URL. The CA used to sign this certificate will be needed by the {agent} clients, not by the {fleet-server} itself.
 
-* The CA certificate (or certificates) used in the certificate chain of your {es} cluster, except if you are sure your {es} certificate is fully trusted publicly.
+* The *CA certificate* (or certificates) associated to your {es} cluster, except if you are sure your {es} certificate is fully trusted publicly.
 
 [discrete]
 [[default-port-assignments-kubernetes]]
@@ -85,16 +75,17 @@ You may need to allow access to these ports. Refer to the following table for de
 
 |===
 | Component communication | Default port
-
-| Elastic Agent → {fleet-server} | 8220
-| Elastic Agent → {es} | 9200
-| Elastic Agent → Logstash | 5044
-| Elastic Agent → {kib} ({fleet}) | 5601
-| {fleet-server} → {kib} ({fleet}) | 5601
+| {agent} → {fleet-server} | 8220
 | {fleet-server} → {es} | 9200
+| {fleet-server} → {kib} (optional, for {fleet} setup) | 5601
+| {agent} → {es} | 9200
+| {agent} → Logstash | 5044
+| {agent} → {kib} (optional, for {fleet} setup) | 5601
 |===
 
-{fleet-server} default port is `8220`, but on Kubernetes you will create a `service` to expose the {fleet-server} deployment. Because of that you will be able to use any other port (for example `443`) without the need of changing the listening port of the {fleet-server}.
+{fleet-server} default port is `8220`, but on Kubernetes you will create a `service` to expose the {fleet-server} deployment. Because of that you will be able to use any other port (for example `443`) without the need of changing the listening port of the {fleet-server} itself.
+(TBD, rewording to something like...
+In Kubernetes environments you can adapt these ports without changing the applications' listening ports, as traffic is managed by Kubernetes services.)
 
 [discrete]
 [[add-fleet-server-kubernetes-add-server]]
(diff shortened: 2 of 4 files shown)


kilfoyle:
@eedugon I read through and this looks great!
Since the PR is in draft mode, should I add comments now or would you like me to wait before doing that?

eedugon:
I still need to add the main kubernetes-related steps and also review all
the content as it was written through a lot of iterations and text
replacements :)

Feel free to review and make comments, in the PR or in private. Now or at a
later stage, whatever you prefer.

A lot of sentences will definitely need improvement as they are currently
more the ideas “I want to express” without thinking much on “how to express
them”, so I’m sure your feedback will help a lot.


Edu González de la Herrán
Senior Technical Writer @ Elastic

*Please note: Emails may be stored and shared to provide a better customer
experience. To learn more about how we use your information, please review
Elastic’s **Privacy Statement. *
<https://www.elastic.co/legal/privacy-statement>

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
+[[add-fleet-server-kubernetes]]
+= Deploy Fleet Server on Kubernetes
+
+[NOTE]
+====
+For ECK users, if your {stack} is orchestrated by ECK, we recommend to orchestrate the {fleet-server} through the operator. That simplifies the process as all needed resources will be configured by ECK.
```
bmorelli25, line 6: I understand what you're going for, but the start of this sentence is repetitive. I would remove the "For ECK users" part.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
++
```
bmorelli25, line 7: It's a bit confusing, but you don't need a `+` here. This actually renders in the output in this instance.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+Refer to {eck-ref}/k8s-elastic-agent.html[Run {agent} on ECK] for more information.
+====
+
+[IMPORTANT]
+====
+This guide assumes certain knowledge about deploying applications on Kubernetes.
```
bmorelli25, line 13: This guide assumes familiarity with deploying applications on Kubernetes. ?

"Certain knowledge" is interesting. I wonder what David thinks 👀

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+* For incoming connections:
+** A *TLS/SSL certificate and key* for the {fleet-server} HTTPS endpoint, used to encrypt the traffic from the {agent}s. This certificate has to be valid for the *{fleet-server} Host URL* that {agent}s use when connecting to the {fleet-server}.
+* Extra TLS/SSL certificates and configuration parameters in case of requiring <<mutual-tls,mutual TLS>> (not covered on this document).
+
+This document walks you through the complete setup process, organized into the following sections:
+
+* <<add-fleet-server-kubernetes-compatibility,Compatibility requirements>>
+* <<add-fleet-server-kubernetes-cert-prereq>>
+* <<add-fleet-server-kubernetes-preparations>>
+* <<add-fleet-server-kubernetes-install>>
```
bmorelli25, line 39: Love it!

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+be running and accessible to your hosts.
+
+You can deploy {fleet-server} on Kubernetes and manage it yourself.
+In this deployment model, you are responsible for high-availability,
+fault-tolerance, and lifecycle management of the {fleet-server}.
+
+To deploy a {fleet-server} on Kubernetes and register it into Fleet you will need the following details:
+
+* The *Policy ID* of a Fleet policy configured with the {fleet-server} integration.
+* A *Service token*, that you can generate following the <<add-fleet-server-kubernetes-preparations>> section, or manually using the {ref}/service-tokens-command.html[elasticsearch-service-tokens] command.
```
bmorelli25, line 26: I would consider removing the links from this bullet point. Why? You don't link out in any of the other bullet points—instead, you use this section to explain what the user needs and why. This bullet point is a bit different; you explain that a service token is needed, but instead of explaining why or what it is, you mention the two methods of creating one. I think it would be better to say something like:

* A *Service token*, used to authenticate Fleet Server with Elasticsearch.

You can then explain the two methods of generating a token in the section below.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+
+. In {kib}, open *Fleet → Settings* and ensure the *Elasticsearch output* that will be used by the {fleet-server} policy is correctly configured, paying special attention to:
++
+** The *hosts* field includes a valid URL that will be reachable by the {fleet-server} Pod.
+** If your {es} cluster uses certificates signed by private or intermediate CAs not publicly trusted, ensure to add the trust information in the *Elasticsearch CA trusted fingerprint* field or in the *advanced configuration* section through `ssl.certificate_authorities` setting. For more details refer to https://elastic.co/guide/en/fleet/current/secure-connections.html#_encrypt_traffic_between_elastic_agents_fleet_server_and_elasticsearch[Secure Connections] documentation.
++
+[IMPORTANT]
+====
+This validation step is critical. The {es} host URL and CA information has to be added *in both the {es} output and the environment variables* provided to the {fleet-server} in a later step. It's a common mistake to ignore the output settings believing that the environment variables will prevail, when the environment variables are only used during the bootstrap of the {fleet-server}.
++
```
bmorelli25, line 145: Another `+` that isn't needed 😄

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+The {fleet-server} certificate is not required when installing the server using the *Quick start* mode, but should always be used for *production* deployments. In *Quick start* mode, the {fleet-server} uses a self-signed certificate and the {agent}s have to be enrolled with the `--insecure` option.
+====
+
+If your organization already uses the {stack}, you may already have a CA certificate that could be used to generate the new cert for the {fleet-server}. If you do not have a CA certificate, you can read more about generating a CA and a server certificate for the {fleet-server} in <<generate-fleet-server-certs>>.
+
+You also need to know and plan in advance the <<fleet-server-hosts-setting, hostname / URL>> that the {agent} clients will use to access the {fleet-server}, because the *hostname* part of the URL needs to be included in the server certificate as an `x.509 Subject Alternative Name (SAN)`. If you plan to make your {fleet-server} accessible through multiple *hostnames* or *FQDNS*, add all of them to the server certificate.
+
+[discrete]
+[[add-fleet-server-kubernetes-cert-outbound]]
+=== {fleet-server} → {es} output outbound traffic flow
```
bmorelli25, line 86: I like this use of ➡️s

bmorelli25 (commented):
I through in a few small grammar things, but generally liking the new flow!

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: Sure! That part together with the `certain knowledge required` was a bit of a draft / placeholder, and I had forgotten to review and update it. I'll do it now.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: Sorry I hadn't finished that sentence... I left only that as a placeholder to complete it at a later stage :)

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: thanks!! :)

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: thank @dedemorton about the `->`s :-D

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: makes sense! and it was redundant actually. thanks!

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+[NOTE]
+====
+If your {es} cluster is on Elastic Cloud or if it uses a certificate signed by a public and known CA you won't need the {es} CA during the setup.
+====
+
+In summary, you need:
+
+* A *server certificate and key*, valid for the {fleet-server} URL. The CA used to sign this certificate will be needed by the {agent} clients, not by the {fleet-server} itself.
+
+* The *CA certificate* (or certificates) associated to your {es} cluster, except if you are sure your {es} certificate is fully trusted publicly.
```
eedugon, line 99: @bmorelli25 : do you think that `summary` is a bit redundant? In theory it's a summary of both traffic flows needs, and in the introduction we also mention both items (cert+key and CA cert of ES).

docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
```diff
@@ -0,0 +1,176 @@
...
+      app: fleet-server
+  template:
+    metadata:
+      labels:
+        app: fleet-server
+    spec:
+      automountServiceAccountToken: false <2>
+      containers:
+      - name: elastic-agent
+        image: docker.elastic.co/beats/elastic-agent:{version}
```
eedugon, line 57: @bmorelli25 , @kilfoyle : Is there any way (`attribute` or hack) to make the `version` to be automatically rendered to the right version? For example in `current`, today it should be rendered as `8.16.1`.

kilfoyle: @eedugon  You found it! That `{version}` attribute will resolve to the current version in the docs. We use it [here](https://github.com/elastic/ingest-docs/blob/1253e201a23639ff38ff31ee5ae93f88733fac69/docs/en/ingest-management/beats-agent-comparison.asciidoc?plain=1#L68), for example.

eedugon: oh! thanks @kilfoyle!
Is it normal then that the {version} is not rendered on previews and in local builds?

because the current preview is still showing it as {version}, and I was a bit scared because of it :)

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: I have updated it, let me know your thoughs @bmorelli25 :)

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: Text updated, let me know how it looks now, please @bmorelli25 :)

docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
bmorelli25: Good question! You're so close. You have to explicitly tell the codeblock to use attribute substitution. To do this, you use `[... subs="attributes"]`. I'll add a comment showing this.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
```diff
@@ -0,0 +1,176 @@
...
+<3> Substitute `PATH_TO_FLEET_SERVER_CERT_KEY` with your local file containing the server TLS key for the {fleet-server}.
++
+If you prefer to obtain a `yaml manifest` of the Secret to create, append `--dry-run=client -o=yaml` to the command and save the output to a file.
+// end::production-secret[]
+
+// ***************************************************
+// ***************************************************
+
+// tag::quickstart-deployment[]
+[source, yaml]
```
bmorelli25, line 38: Enable attribute substitution

```suggestion
["source","yaml",subs="attributes"]
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
bmorelli25: https://github.com/elastic/ingest-docs/pull/1518/files#r1878558170

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
bmorelli25: much better!

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
bmorelli25: Hmmm. Yes and no. I can see a case for keeping this and for removing it. I don't feel strongly either way.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
eedugon: thanks!!

docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
eedugon: all clear now, thanks!

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+[[add-fleet-server-kubernetes-cert-prereq]]
+== {fleet-server} and SSL/TLS certificates considerations
+
+This section shows the minimum requirements in terms of Transport Layer Security (TLS) certificates for the {fleet-server}, assuming no mutual TLS (mTLS) is needed. Refer to <<tls-overview>> and <<mutual-tls>> for more information about the configuration needs of both approaches.
+
+There are two main and differentiated traffic flows associated with the {fleet-server}, where each of them has different TLS needs:
+
+[discrete]
+[[add-fleet-server-kubernetes-cert-inbound]]
+=== {agent} → {fleet-server} inbound traffic flow
```
laraMorenoIgle, line 71:
```suggestion
=== [{agent} → {fleet-server}] inbound traffic flow
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+A {fleet-server} certificate is not required when installing the server using the *Quick start* mode, but should always be used for *production* deployments. In *Quick start* mode, the {fleet-server} uses a self-signed certificate and the {agent}s have to be enrolled with the `--insecure` option.
+====
+
+If your organization already uses the {stack}, you may already have a CA certificate that could be used to generate the new cert for the {fleet-server}. If you do not have a CA certificate, you can read more about generating a CA and a server certificate for the {fleet-server} in <<generate-fleet-server-certs>>.
+
+You also need to know and plan in advance the <<fleet-server-hosts-setting, hostname / URL>> that the {agent} clients will use to access the {fleet-server}, because the *hostname* part of the URL needs to be included in the server certificate as an `x.509 Subject Alternative Name (SAN)`. If you plan to make your {fleet-server} accessible through *multiple hostnames* or *FQDNs*, add all of them to the server certificate.
+
+[discrete]
+[[add-fleet-server-kubernetes-cert-outbound]]
+=== {fleet-server} → {es} output outbound traffic flow
```
laraMorenoIgle, line 86:
```suggestion
=== [{fleet-server} → {es} output] outbound traffic flow
```

laraMorenoIgle (commented):
As we have talked in private, I suggest:
- Add some context for quickstart and Production mode
- Add a link to the environment variables available for fleet server https://www.elastic.co/guide/en/fleet/current/agent-environment-variables.html

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+To deploy a {fleet-server} on Kubernetes and register it into Fleet you will need the following details:
+
+* The *Policy ID* of a Fleet policy configured with the {fleet-server} integration.
+* A *Service token*, used to authenticate Fleet Server with Elasticsearch.
+* For outgoing traffic:
+** The *{es} endpoint URL* where the {fleet-server} should connect to, configured also in the {es} output associated to the policy.
+** When a private or intermediate Certificate Authority (CA) is used to sign the {es} certificate, the *{es} CA file* or the *CA fingerprint*, configured also in the {es} output associated to the policy.
+* For incoming connections:
+** A *TLS/SSL certificate and key* for the {fleet-server} HTTPS endpoint, used to encrypt the traffic from the {agent}s. This certificate has to be valid for the *{fleet-server} Host URL* that {agent}s use when connecting to the {fleet-server}.
+* Extra TLS/SSL certificates and configuration parameters in case of requiring <<mutual-tls,mutual TLS>> (not covered on this document).
```
kilfoyle, line 32:
```suggestion
* Extra TLS/SSL certificates and configuration parameters in case of requiring <<mutual-tls,mutual TLS>> (not covered in this document).
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+
+[discrete]
+[[add-fleet-server-kubernetes-compatibility]]
+= Compatibility
+
+{fleet-server} is compatible with the following Elastic products:
+
+* {stack} 7.13 or later.
+** For version compatibility: {es} >= {fleet-server} >= {agent} (except for
+bugfix releases)
```
kilfoyle, line 49:
```suggestion
** For version compatibility: {es} >= {fleet-server} >= {agent} (except for
bugfix releases).
```

kilfoyle, line 49:
```suggestion
** For version compatibility, {es} must be at the same or a later version than {fleet-server}, and {fleet-server} needs to be at the same or a later version than {agent} (not including patch releases). 
```
Personally, I'd probably write this out rather than use the `>=` signs, but I think either one is just fine.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+** {kib} should be on the same minor version as {es}.
+
+[discrete]
+[[add-fleet-server-kubernetes-prereq]]
+= Prerequisites
+
+Before deploying the {fleet-server}, you need to:
```
kilfoyle, line 56:
```suggestion
Before deploying {fleet-server}, you need to:
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+This document walks you through the complete setup process, organized into the following sections:
+
+* <<add-fleet-server-kubernetes-compatibility,Compatibility requirements>>
+* <<add-fleet-server-kubernetes-cert-prereq>>
+* <<add-fleet-server-kubernetes-preparations>>
+* <<add-fleet-server-kubernetes-install>>
+
+[discrete]
+[[add-fleet-server-kubernetes-compatibility]]
+= Compatibility
```
kilfoyle, line 43: I see that the document structure is like this:

```
= Deploy Fleet Server on Kubernetes

= Compatibility

= Prerequisites
== {fleet-server} and SSL/TLS certificates considerations
=== [{agent} → {fleet-server}] inbound traffic flow
=== [{fleet-server} → {es} output] outbound traffic flow
== Default port assignments
== Fleet preparations
== Fleet Server installation
=== Installation Steps

= Troubleshooting
== Common Problems

= Next steps
```

So all of the preparations and install steps are part of the "Prerequisites" section in the hierarchy.

I think what you want is probably more like this:

```
= Deploy Fleet Server on Kubernetes

== Compatibility

== Prerequisites
=== {fleet-server} and SSL/TLS certificates considerations
==== [{agent} → {fleet-server}] inbound traffic flow
==== [{fleet-server} → {es} output] outbound traffic flow
=== Default port assignments

== Fleet preparations

== Fleet Server installation
=== Installation Steps

== Troubleshooting
=== Common Problems

== Next steps
```

Also, I would include all of the second level items (Compatibility, Prerequisites, Fleet preparations, Fleet Server installation, Troubleshooting, Next steps) in the list of sections at line 36-39. It's nice that you put that list in because it'll help users to understand the document structure.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+* Prepare the SSL/TLS configuration, server certificate, <<fleet-server-hosts-setting>>, and needed Certificate Authorities (CAs).
+* Ensure components have access to the ports needed for communication.
+
+[discrete]
+[[add-fleet-server-kubernetes-cert-prereq]]
+== {fleet-server} and SSL/TLS certificates considerations
+
+This section shows the minimum requirements in terms of Transport Layer Security (TLS) certificates for the {fleet-server}, assuming no mutual TLS (mTLS) is needed. Refer to <<tls-overview>> and <<mutual-tls>> for more information about the configuration needs of both approaches.
+
+There are two main and differentiated traffic flows associated with the {fleet-server}, where each of them has different TLS needs:
```
kilfoyle, line 67:
```suggestion
There are two main traffic flows available for {fleet-server}, each with different TLS needs:
```
Just trying to simplify it a bit. :-)

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+
+[discrete]
+[[add-fleet-server-kubernetes-cert-inbound]]
+=== [{agent} → {fleet-server}] inbound traffic flow
+
+In this flow the {fleet-server} acts as the server and the {agent} acts as the client, therefore the {fleet-server} requires a TLS certificate and key, and the {agent} will need to trust the CA certificate used to sign the {fleet-server} certificate.
```
kilfoyle, line 73:
```suggestion
In this flow {fleet-server} acts as the server and the {agent} acts as the client. Therefore, {fleet-server} requires a TLS certificate and key, and {agent} will need to trust the CA certificate used to sign the {fleet-server} certificate.
```
I think this reads better as two sentences. Also, we usually say "Fleet Server" rather than "a/the Fleet Server".

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+
+You also need to know and plan in advance the <<fleet-server-hosts-setting, hostname / URL>> that the {agent} clients will use to access the {fleet-server}, because the *hostname* part of the URL needs to be included in the server certificate as an `x.509 Subject Alternative Name (SAN)`. If you plan to make your {fleet-server} accessible through *multiple hostnames* or *FQDNs*, add all of them to the server certificate.
+
+[discrete]
+[[add-fleet-server-kubernetes-cert-outbound]]
+=== [{fleet-server} → {es} output] outbound traffic flow
+
+In this flow the {fleet-server} acts as the client and {es} acts as the HTTPS server. For the communication to succeed, the {fleet-server} needs to trust the CA certificate used to sign the {es} certificate. If your {es} uses certificates signed by a corporate CA or multiple intermediate CAs you will need to use them during the {fleet-server} setup.
```
kilfoyle, line 88:
```suggestion
In this flow, {fleet-server} acts as the client and {es} acts as the HTTPS server. For the communication to succeed, {fleet-server} needs to trust the CA certificate used to sign the {es} certificate. If your {es} cluster uses certificates signed by a corporate CA or multiple intermediate CAs you will need to use them during the {fleet-server} setup.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+
+[NOTE]
+====
+If your {es} cluster is on Elastic Cloud or if it uses a certificate signed by a public and known CA you won't need the {es} CA during the setup.
```
kilfoyle, line 92:
```suggestion
If your {es} cluster is on Elastic Cloud or if it uses a certificate signed by a public and known CA, you won't need the {es} CA during the setup.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+== Fleet preparations
+
+[TIP]
+====
+If you already have a Fleet policy with the {fleet-server} integration, you know its ID, and you know how to generate an {ref}/service-tokens-command.html[{es} service token] for the {fleet-server}, skip directly to <<add-fleet-server-kubernetes-install>>.
+
+Also note that the `service token` required by the {fleet-server} is different from the `enrollment tokens` used by {agent}s to enroll to Fleet.
+====
+
+. In {kib}, open *Fleet → Settings* and ensure the *Elasticsearch output* that will be used by the {fleet-server} policy is correctly configured, paying special attention to:
```
kilfoyle, line 137:
```suggestion
. In {kib}, open *Fleet → Settings* and ensure the *Elasticsearch output* that will be used by the {fleet-server} policy is correctly configured, paying special attention that:
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
++
+** The *hosts* field includes a valid URL that will be reachable by the {fleet-server} Pod.
+** If your {es} cluster uses certificates signed by private or intermediate CAs not publicly trusted, ensure to add the trust information in the *Elasticsearch CA trusted fingerprint* field or in the *advanced configuration* section through `ssl.certificate_authorities` setting. For more details refer to https://elastic.co/guide/en/fleet/current/secure-connections.html#_encrypt_traffic_between_elastic_agents_fleet_server_and_elasticsearch[Secure Connections] documentation.
```
kilfoyle, line 140:
```suggestion
** If your {es} cluster uses certificates signed by private or intermediate CAs not publicly trusted, you have added the trust information in the *Elasticsearch CA trusted fingerprint* field or in the *advanced configuration* section through the `ssl.certificate_authorities` setting. For details, refer to the https://elastic.co/guide/en/fleet/current/secure-connections.html#_encrypt_traffic_between_elastic_agents_fleet_server_and_elasticsearch[Secure Connections] documentation.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
++
+[IMPORTANT]
+====
+This validation step is critical. The {es} host URL and CA information has to be added *in both the {es} output and the environment variables* provided to the {fleet-server} in a later step. It's a common mistake to ignore the output settings believing that the environment variables will prevail, when the environment variables are only used during the bootstrap of the {fleet-server}.
```
kilfoyle, line 144: Maybe "in a later step" should be removed? It sounds to me a bit like the reader will add these configs as part of steps described later on in the document.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+
+Consider the creation of a dedicated *{es} output* for the {fleet-server} if the URL needed by the {fleet-server} to reach {es} differs from the URL needed by other clients.
```
kilfoyle, line 146:
```suggestion
If the URL that {fleet-server} will use to access {es} is different from the {es} URL used by other clients, you may want to create a dedicated *{es} output* for {fleet-server}.
```
I think I'd reverse the order to make it a bit more clear, but the original is okay too.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+====
+
+. From *Fleet → Agent Policies*, select *Create agent policy* to create a policy for the {fleet-server}:
```
kilfoyle, line 149:
```suggestion
. Go to *Fleet → Agent Policies* and select *Create agent policy* to create a policy for the {fleet-server}:
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
++
+** Set a *name* to the policy, for example `Fleet Server Policy Kubernetes`
```
kilfoyle, line 151:
```suggestion
** Set a *name* for the policy, for example `Fleet Server Policy Kubernetes`.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+** Do *not* select the option *collect system logs and metrics*. This option would add the `system integration` to the policy, and considering the {fleet-server} will run as a Kubernetes Pod without any visibility to the Kubernetes node, there won't be any real system to monitor.
```
kilfoyle, line 152:
```suggestion
** Do *not* select the option *Collect system logs and metrics*. This option adds the System integration to the {agent} policy. Since {fleet-server} will run as a Kubernetes Pod without any visibility to the Kubernetes node, there won't be a system to monitor.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+** Select the **output** that the {fleet-server} needs to use to contact {es}. This should be the output that was verified in the previous step.
```
kilfoyle, line 153:
```suggestion
** Select the **output** that the {fleet-server} needs to use to contact {es}. This should be the output that you verified in the previous step.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+** If desired, add an **inactive agent unenrollment timeout** to automatically unenroll and invalidate API Keys after being inactive. This is useful on Kubernetes environments as the {fleet-server} Pods are ephemeral, and after restarts new {agent}s will appear in Fleet UI.
```
kilfoyle, line 154:
```suggestion
** Optionally, you can add an **inactive agent unenrollment timeout** to automatically unenroll and invalidate API keys after they become inactive. This is useful on Kubernetes environments as the {fleet-server} Pods are ephemeral, and after restarts new {agent}s will appear in Fleet UI.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+
+. Open the created policy, and from the *Integrations* tab select *Add integration*:
++
+** Search and select the Fleet Server integration
```
kilfoyle, line 158:
```suggestion
** Search for and select the {fleet-server} integration.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+** Select *Add Fleet Server* to add the integration to the policy.
```
kilfoyle, line 159:
```suggestion
** Select *Add Fleet Server* to add the integration to the {agent} policy.
```
I usually specify "{agent} policy" since there are also integration policies.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
++
+At this point you can configure the integration settings per <<fleet-server-scalability>>.
+** When done, select on *Save and continue*. Do not add any {agent} at this stage.
```
kilfoyle, line 162:
```suggestion
** When done, select *Save and continue*. Do not add any {agent} at this stage.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+
+. Open the configured policy, which now includes the Fleet Server integration, and select *Actions* → *Add Fleet Server*. In the next dialog:
```
kilfoyle, line 164:
```suggestion
. Open the configured policy, which now includes the {fleet-server} integration, and select *Actions* → *Add Fleet Server*. In the next dialog:
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
kilfoyle: The convention for docs is that any field name marked in bold should exactly match the field name in the UI. However, I don't see one named "output". Should it be `**Output for integrations**`?

Screenshot 2024-12-11 at 2 51 12 PM

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
++
+* Confirm that the *policy for Fleet Server* is properly selected.
```
kilfoyle, line 166: Just checking: Is `policy for Fleet Server` the string exactly as it appears in the UI?

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+* *Choose a deployment mode for security*:
+** If you select *Quick start*, the {fleet-server} generates a `self-signed` TLS certificate, and subsequent agents should be enrolled using the `--insecure` flag.
+** If you select *Production*, you provide a certificate and key to the {fleet-server} during the deployment, and subsequent agents will need to trust the certificate's CA.
+* Add your *Fleet Server Host* information:
+** This will be the URL used by the clients ({agent}s) to connect to the {fleet-server}. In *Production* mode, the {fleet-server} certificate must include the hostname part of the URL as an `x509 SAN`.
```
kilfoyle, line 171:
```suggestion
** This is the URL used by the clients ({agent}s) to connect to the {fleet-server}. In *Production* mode, the {fleet-server} certificate must include the hostname part of the URL as an `x509 SAN`.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+** On Kubernetes environments this could be the name of the `Kubernetes service` or reverse proxy that exposes the {fleet-server} Pods.
+** This setting is not really used by the {fleet-server} itself, so don't worry much about it at this stage because you will be able to change it later on under *Fleet → Settings*.
```
kilfoyle, line 173:
```suggestion
** This setting is not used by the {fleet-server} itself, so don't worry about it at this stage because you will be able to change it later on under *Fleet → Settings*.
```
I'd omit `really` if the setting isn't used by Fleet Server.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+* Select **generate service token** to create a token for the {fleet-server}.
+* From *Install Fleet Server to a centralized host → Linux*, take note of the values of the following settings that will be needed for the {fleet-server} installation:
+** Service token(under `--fleet-server-service-token` parameter).
```
kilfoyle, line 176:
```suggestion
** Service token (under `--fleet-server-service-token` parameter).
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+** Fleet policy ID (under `--fleet-server-policy` parameter).
+** {es} URL (under `--fleet-server-es` parameter).
```
kilfoyle, line 178:
```suggestion
** Service token (below the `--fleet-server-service-token` parameter).
** Fleet policy ID (below the `--fleet-server-policy` parameter).
** {es} URL (below the `--fleet-server-es` parameter).
```
To me, `below` sounds a bit better than `under`, but I'm not sure why. I also added a missing space.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+
+. Keep {kib} browser window opened and continue with the <<add-fleet-server-kubernetes-install>>.
```
kilfoyle, line 180:
```suggestion
. Keep the {kib} browser window open and continue with the <<add-fleet-server-kubernetes-install>>.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
++
+When {fleet-server} installation succeeds, the *Confirm Connection* dialog will show *Connected*.
```
kilfoyle, line 182:
```suggestion
When the {fleet-server} installation has succeeded, the *Confirm Connection* UI will show *Connected*.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+
+[discrete]
+[[add-fleet-server-kubernetes-install]]
+== Fleet Server installation
+
+To deploy {fleet-server} on Kubernetes and register it into Fleet you need the following details:
```
kilfoyle, line 188:
```suggestion
To deploy {fleet-server} on Kubernetes and enroll it into {fleet} you need the following details:
```
We use "enroll" for Agents, so I think for Fleet Server that term might be better than "register".

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+
+* *Policy ID* of the Fleet policy configured with the {fleet-server} integration.
```
kilfoyle, line 190: Should it be "Fleet Server policy" rather than "Fleet policy"?
Also, either way we should use an attribute, either `{fleet}` or `{fleet-server}`.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+* *Service token*, that you can generate following the <<add-fleet-server-kubernetes-preparations>> or manually using the {ref}/service-tokens-command.html[{es}-service-tokens command].
+* *{es} endpoint URL*, configured also in the {es} output associated to the policy.
+* *{es} CA certificate file*, configured also in the {es} output associated to the policy.
```
kilfoyle, line 193: I'd probably remove `also` in these lines. Without it, it reads to me as "You'll need this setting, and here's where you can find it".

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+
+If you followed the <<add-fleet-server-kubernetes-cert-prereq>> and <<add-fleet-server-kubernetes-preparations>> you should have everything ready to proceed with the {fleet-server} installation.
+
+The suggested deployment method for the {fleet-server} consists of:
+
+* A basic Kubernetes `Deployment` manifest that relies on two `Secrets` for its configuration.
+* A `Secret` named `fleet-server-config` with the values for different environment variables.
+* A `Secret` named `fleet-server-certs` with the certificate files and the {es} CA file needed for the installation.
+
+Adapt and change the suggested manifests and deployment strategy to your needs, ensuring you feed the {fleet-server} with the needed configuration and certificates. For example, you can customize:
```
kilfoyle, line 204: Just wanna say that I love the word "feed" here. :-)

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+
+* Scheduling configuration, such as `affinity rules` or `tolerations`, if needed in your environment.
+* CPU and memory resources configuration.
+* Replicas for the Deployment (defaults to 1).
```
kilfoyle, line 208:
```suggestion
* Replicas for the Deployment (defaults to `1`).
```
I'd use backticks since it's a parameter value, but not a big deal either way.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
++
+[source, shell]
+------------------------------------------------------------
+kubectl create secret generic fleet-server-config \
+--from-literal=elastic_endpoint='YOUR_ELASTICSEARCH_HOST_URL' \ <1>
+--from-literal=elastic_service_token='YOUR_SERVICE_TOKEN' \ <2>
+--from-literal=fleet_policy_id='YOUR_POLICY_ID' <3>
+------------------------------------------------------------
+<1> The URL of your {es} host. For example `'https://monitoring-es-http.default.svc:9200'`.
+<2> The service token provided by {kib} in Fleet UI.
```
kilfoyle, line 226:
```suggestion
<2> The service token provided by {kib} in the Fleet UI.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+====
+
+. Check the {fleet-server} Pod logs for errors and confirm in {kib} that the {fleet-server} agent appears as `Connected` and `Healthy` in *{kib} → Fleet*.
++
+[source, shell]
+------------------------------------------------------------
+kubectl logs fleet-server-xxxxx
+------------------------------------------------------------
++
+It could take a while until the {fleet-server} fully starts. If you left the {kib} browser window opened during <<add-fleet-server-kubernetes-preparations>> it will show *Connected* in case everything goes well.
```
kilfoyle, line 336:
```suggestion
It can take a couple of minutes for {fleet-server} to fully start. If you left the {kib} browser window open during <<add-fleet-server-kubernetes-preparations>> it will show *Connected* when everything has gone well.
```
I'd say "a couple of minutes" just so people know how long they should wait in case Fleet Server doesn't enroll successfully, but please change if that's not the right period of time.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
++
+Do not continue with the next steps if the {fleet-server} is not completely up and healthy and appears in Fleet UI.
```
kilfoyle, line 338:
```suggestion
Do not continue with the next steps if the {fleet-server} is not completely up and healthy, and appearing in {fleet} UI.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+
+. Expose the {fleet-server} to {agent}s through the *Fleet Server Host*:
++
+This may include the creation of a Kubernetes `service`, an `ingress` resource, or create DNS registers for FQDNs resolution. There are multiple ways to expose applications in Kubernetes.
```
kilfoyle, line 342:
```suggestion
This may include the creation of a Kubernetes `service`, an `ingress` resource, or DNS registers for FQDNs resolution. There are multiple ways to expose applications in Kubernetes.
```
`creation` and `create` are different tenses. Maybe we can just omit the second one?

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
++
+The following shows an example to expose the {fleet-server} with a `LoadBalancer` service, which will make the {fleet-server} available for both internal and external clients. Consider creating a `ClusterIP` instead of a `LoadBalancer` service if you only need to expose the {fleet-server} to {agent}s running in the same Kubernetes cluster.
```
kilfoyle, line 344:
```suggestion
The following example exposes the {fleet-server} with a `LoadBalancer` service, which will make the {fleet-server} available for both internal and external clients. Consider creating a `ClusterIP` instead of a `LoadBalancer` service if you only need to expose the {fleet-server} to {agent}s running in the same Kubernetes cluster.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+------------------------------------------------------------
+kubectl expose deployment fleet-server --name fleet-svc --type LoadBalancer --port 443 --target-port 8220
+------------------------------------------------------------
++
+The previous example creates a service named `fleet-svc`, listening on port `443` and forwarding the traffic to the `fleet-server` deployment's Pods on port `8220`. The listening `--port` of the service can be customized, but the `--target-port` must remain on the default port (`8220`), because it's the port used by the {fleet-server} application.
++
+The URL(s) that the clients would use to access the {fleet-server} in this example could be:
++
+* For {agent}s within the Kubernetes cluster: `https://fleet-svc.<namespace>.svc`.
+* For external {agent}s: access through whatever FQDN added on the DNS server pointing to the public IP of the `LoadBalancer`.
```
kilfoyle, line 356:
```suggestion
* For external {agent}s: access through the FQDN that was added on the DNS server pointing to the public IP of the `LoadBalancer`.
```
Just a small rephrasing.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
++
+If your environment requires the {fleet-server} to be reachable through multiple hostnames / URLs, you can create multiple *Fleet Server Hosts* in *Fleet → Setting*, and create different policies for different groups of agents.
```
kilfoyle, line 358:
```suggestion
If your environment requires the {fleet-server} to be reachable through multiple hostnames or URLs, you can create multiple *Fleet Server Hosts* in *Fleet → Settings*, and create different policies for different groups of agents.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+[[add-fleet-server-kubernetes-troubleshoot]]
+= Troubleshooting
+
+[discrete]
+[[add-fleet-server-kubernetes-troubleshoot-common]]
+== Common Problems
+
+* {fleet-server} trying to access {es} at `localhost:9200` even if `FLEET_SERVER_ELASTICSEARCH_HOST` environment variable was properly set:
++
+This occurs when the `output` of the policy associated to the {fleet-server} is not correctly configured.
```
kilfoyle, line 372:
```suggestion
* {fleet-server} is trying to access {es} at `localhost:9200` even though the `FLEET_SERVER_ELASTICSEARCH_HOST` environment variable is properly set.
+
This problem occurs when the `output` of the policy associated to the {fleet-server} is not correctly configured.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+
+* TLS certificate trust issues even when `ELASTICSEARCH_CA` environment variable is properly set during deployment:
++
+This occurs when the `output` of the policy associated to the {fleet-server} is not correctly configured. Add the *CA certificate* or *CA trusted fingerprint* to the {es} output associated to the {fleet-server} policy.
```
kilfoyle, line 376:
```suggestion
* TLS certificate trust issues occur even when the `ELASTICSEARCH_CA` environment variable is properly set during deployment.
+
This problem occurs when the `output` of the policy associated to the {fleet-server} is not correctly configured. Add the *CA certificate* or *CA trusted fingerprint* to the {es} output associated to the {fleet-server} policy.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+Remember that in *Production* mode, the *hostnames* used to access the {fleet-server} must be part of the server certificate as `x.509 Subject Alternative Names`.
+
+[discrete]
+[[add-fleet-server-kubernetes-troubleshoot]]
+= Troubleshooting
+
+[discrete]
+[[add-fleet-server-kubernetes-troubleshoot-common]]
+== Common Problems
+
```
kilfoyle, line 369:
```suggestion

The following issues may occur when {fleet-server} settings are missing or configured incorrectly:

```
I think it would be nice to have a line introducing this section, especially since we may add more items over time.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+If you're unable to add a {fleet}-managed agent, click the **Agents** tab
+and confirm that the agent running {fleet-server} is healthy.
+
+[discrete]
+[[add-fleet-server-kubernetes-next]]
+= Next steps
+
+Now you're ready to add {agent}s to your host systems.
+To learn how, see <<install-fleet-managed-elastic-agent>>, or <<running-on-kubernetes-managed-by-fleet>> if your {agent}s will also run on Kubernetes.
```
kilfoyle, line 390:
```suggestion
To learn how, refer to <<install-fleet-managed-elastic-agent>>, or <<running-on-kubernetes-managed-by-fleet>> if your {agent}s will also run on Kubernetes.
```

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
```diff
@@ -0,0 +1,392 @@
...
+
+When connecting {agent}s to this {fleet-server}, remember to use `--insecure` flag if the *quick start* mode was used, or to provide to the {agent}s the CA certificate associated to the {fleet-server} certificate if *production* mode was used.
```
kilfoyle, line 392:
```suggestion
When connecting {agent}s to {fleet-server}, remember to use the `--insecure` flag if the *quick start* mode was used, or to provide to the {agent}s the CA certificate associated to the {fleet-server} certificate if *production* mode was used.
```

kilfoyle (approved):
@eedugon  this is really nice and our users are going to find it super helpful! Thanks a lot for writing this up.

I added several comments mostly with small rephrasings, just for consistency with our other docs, but overall I think it's in great shape.

I'll approve it now but I'm happy to give it a re-read anytime, if you like.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: Ok! we will keep this for the moment and review at a later stage ;)

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: Totally agreed. As mentioned in the previous comment, the compatibility section was copied from the other docs about fleet server installation, so we should probably fix that everywhere.

eedugon: The compatibility section was copied from the other docs about fleet server installation, so we should probably fix that everywhere.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: What about `There are two main traffic flows related (to or with) {fleet-server}, each with different TLS needs`?

The word `available` for traffic flows doesn't sound accurate to me, as a traffic flow is not something available or unavailable, but more something that exists and happens or doesn't exist. What do you think? :)

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: > Without it, it reads to me as "You'll need this setting, and here's where you can find it".

mmmm no, the objective is to remember the users that they need the ES endpoint URL configured in BOTH places (same for the CA certificate). We are not telling the users from where to obtain the info (except for the service-token).

And actually, the ES outuput in Fleet UI will appear by default with an "invalid" URL like `http://localhost:9200`, and the CA information will be empty.

What I try to remind the users (maybe too many times) is that both the CA and the URL need to be added in two different places.

But it's true that this isn't intuitive in the current doc text.... what about this?

> *{es} endpoint URL*, configured in both the {es} output associated to the policy and in the Fleet Server as an environment variable.
> *{es} CA certificate file*, configured in both the {es} output associated to the policy and in the Fleet Server.

Let me know your thoughts @kilfoyle, and thanks! ;)

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: I'll review the attributes, thanks!

In this case IMO it should be a Fleet policy (configured with Fleet Server integration).
Fleet Server policy makes total sense conceptually speaking, but there's no such a thing, you can name a policy `Fleet Server policy` or `fleet-server-policy` but it's still a policy.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: This is a very good point.

I didn't want to use the word `enroll` on purpose, to avoid users mixing the enrollment token needed by Elastic Agents to enroll into Fleet (the normal Elastic Agents `enroll` to Fleet via the Fleet Server), with the `service token` that the Fleet Server needs to join Fleet.

I don't know what exactly the Fleet Server does when joining Fleet with the service token, I don't know if it's technically similar to an enrollment, probably yes :)

I've reviewed other Fleet Server install docs and clearly we consider it `enrolled` to Fleet, so I'm applying the change. Thanks a lot for pointing this out!

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: what about `specified by` or `provided after`? :)

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: Yes, it appears exactly like that, as part of the sentence Select a policy for the Fleet Server:

image

I'm opened to suggestions here :)

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: Thanks! I'll review all attributes before merging, as I might have missed more :)

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: I like the change proposal, thanks!

I've one concern about the `Since` word, as I learned short time ago through another review that we should use `because` instead.

The reason is that `since` should only be used when it really means `since` a given period of time or it expresses a reason in the past.

But not totally sure... @bmorelli25 , what do you think here? :)

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: Thanks!
I've left it as:
```
= Deploy Fleet Server on Kubernetes
== Compatibility
== Prerequisites
=== {fleet-server} and SSL/TLS certificates considerations
==== [{agent} → {fleet-server}] inbound traffic flow
==== [{fleet-server} → {es} output] outbound traffic flow
=== Default port assignments
== Add {fleet-server}
=== Fleet preparations
=== Fleet Server installation
==== Installation overview
==== Installation Steps
== Troubleshoot {fleet-server}
=== Common Problems
=== Other issues
== Next steps
```

And updated the lines 36-39.
Let me know your thoughts @kilfoyle

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: done! :)

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
eedugon: updated already, let me know your thoughts. I finally wrote:

`There are two main traffic flows for {fleet-server}, each with different TLS requirements:`

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
kilfoyle: Ah, indeed! "Because" would certainly be better than "Since", *because* the latter is ambiguous. Thanks for catching that Edu!

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
kilfoyle: I can make that change in the other docs.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
kilfoyle: Looks good. Thanks!

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
kilfoyle: Nice! I think it's perfect now.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
kilfoyle: Thanks for confirming. It matches the UI so I think your sentence is perfect as is.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
kilfoyle: I really like `specified by`.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
kilfoyle: Sorry, this is my mistake. I was getting my terms confused. I think Fleet policy is okay as you've explained.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
kilfoyle: Thanks for explaining Edu! I really like the the proposed text.

This seems 100% clear:

> *{es} endpoint URL*, configured in both the {es} output associated to the policy and in the Fleet Server as an environment variable.
> *{es} CA certificate file*, configured in both the {es} output associated to the policy and in the Fleet Server.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes.asciidoc
kilfoyle: Here's a PR to make the same change in the other places in the Fleet & Agent docs: https://github.com/elastic/ingest-docs/pull/1537

eedugon:
@kilfoyle, after reviewing and applying the latest updates to the doc together I'd like devs to review the content from the technical side also.

@swiatekm : would you be able to take a look and share your thoughts?

I'd like your comments around the specific proposal to install the Fleet Server on Kubernetes. We are suggesting a single 
`Deployment` backed up by 2 secrets, one for env vars, and another one for the CA certificate of Elasticsearch. The manifest doesn't need any RBAC and we disable the token to talk to Kubernetes API because the Fleet Server doesn't need it.

If you believe there's a quicker or better approach let us know. I haven't tried the automatic Kibana Setup directly by the Agent (using *KIBANA_* env vars) because I've never been able to make it working (that functionality is not properly documented either).

cc: @nimarezainia : we would also like you to review this new document :)

swiatekm:
@eedugon I did a first pass review and everything looks good to me. Before approving, I'm also going to actually follow the guide as well in my test environment to see if I spot any sharp edges that way.

One note for the future is that the CA setup seems like the most complex part of the guide. While we can't do much about this when it comes to ES itself, for the connection between Fleet Server and agent, we should be able to have the Fleet Server Helm Chart generate the necessary certs, or, even better, have cert-manager do it if it's available. WDYT @pkoutsovasilis ?

swiatekm (commented):
Found a few issues, comments below.

eedugon:
Thanks a lot!
I will work on that.
Thanks for raising the issue when using the cert and key. I’ve realized
that I didn’t test that part!!

pkoutsovasilis:
sorry for the delay guys 👋 I will review this shortly 🙂

docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
```diff
@@ -0,0 +1,176 @@
...
+
+// tag::production-secret[]
+The following command assumes you have the {es} CA file, the {fleet-server} certificate, and the {fleet-server} key available in local files.
++
+[source, shell]
+------------------------------------------------------------
+kubectl create secret generic fleet-server-certs \
+  --from-file=es-ca.crt=<PATH_TO_ES_CA_CERT_FILE> \ <1>
+  --from-file=fleet-server.crt=<PATH_TO_FLEET_SERVER_CERT> \ <2>
+  --from-file=fleet-server.key=<PATH_TO_FLEET_SERVER_CERT_KEY> <3>
```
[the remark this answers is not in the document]
eedugon: Solved, thanks a lot for the feedback @swiatekm

docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
```diff
@@ -0,0 +1,176 @@
...
+            valueFrom:
+              secretKeyRef:
+                name: fleet-server-config
+                key: fleet_policy_id
+          - name: ELASTICSEARCH_CA
+            value: /mnt/certs/es-ca.crt
+          - name: FLEET_SERVER_CERT
+            value: /mnt/certs/fleet-server.crt
+          - name: FLEET_SERVER_KEY
+            value: /mnt/certs/fleet-server.key
```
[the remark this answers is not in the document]
eedugon: Thanks a lot! I have now fully tested the production mode and applied multiple changes in the doc due to this.

First of all because it's not intuitive to need your access your own URL and use your own CA for validation purposes when starting a SSL server.

eedugon:
@kilfoyle : Would you please perform a final review of the doc?

Due to some important findings by @swiatekm I had to apply some changes to the doc content to ensure `Production` mode fully works and is fully understood by users.

@swiatekm and @pkoutsovasilis : I'd also appreciate your review here also if you have some time, as it would be great to publish this document before the documentation `freeze` we will have soon.

I think this document, even if it's not perfect, it's going to provide a lot of help to users that have already tried in the past to run the fleet server on Kubernetes.

Of course if in the future we simplify things by running directly the `fleet-server binary` without the need of an `agent enrolling to itself` or if we create a helm chart to handle all this we will be able to update the document.

docs/en/ingest-management/fleet/add-fleet-server-kubernetes-content.asciidoc
```diff
@@ -0,0 +1,209 @@
...
+            valueFrom:
+              secretKeyRef:
+                name: fleet-server-config
+                key: elastic_service_token
+          - name: FLEET_SERVER_POLICY_ID
+            valueFrom:
+              secretKeyRef:
+                name: fleet-server-config
+                key: fleet_policy_id
+          - name: ELASTICSEARCH_CA
```
pkoutsovasilis, line 103: qq @eedugon is this actually needed for the quick-start deployment? 🙂

eedugon: > is this actually needed for the quick-start deployment?
@pkoutsovasilis , thanks a lot for looking into this!

Yes, I think the `ELASTICSEARCH_CA` is needed in both quick start and production modes / cases.

Quick start mode only affects the `HTTPS server` part of the `fleet server`. In quickstart mode the fleet server creates a `self-signed` certificate and the clients need to be enrolled with `--insecure` flag, but that doesn't affect the traffic flows with Elasticsearch.

At least that's how I understand this, and in my tests I always need to set the ELASTICSEARCH_CA of my self-managed elasticsearch cluster when booting a Fleet Server in quickstart mode.

eedugon: well, of course if the Elasticsearch certificate is signed by a well known and public CA then `ELASTICSEARCH_CA` won't be needed, but that will also be the same no matter `quick start` / `production`.

I believe I have also mentioned that in the doc in a couple of places.

bmorelli25 (approved):
This is looking great! 🚀

Issue #1547, opened 2024-12-16
Update MSI install docs for Powershell


kilfoyle:
From Bill:

For this MSI doc https://www.elastic.co/guide/en/fleet/current/install-agent-msi.html

If the user runs the command via Powershell we need them to use a slightly different command:

`./elastic-agent-8.15.0-windows-x86_64.msi --% INSTALLARGS="--url=url --enrollment-token=token=="`

with the executable targeted with `./` and `--%` after the MSI to prevent Powershell from adding quotes around INSTALLARGS due to the space between url and enrollment token

PR #1548, opened 2024-12-16
Add PowerShell variant to MSI install docs

kilfoyle:
This updates the [MSI install page](https://www.elastic.co/guide/en/fleet/current/install-agent-msi.html) since the PowerShell version of the install command for Elastic Agent is a bit different.

Closes: https://github.com/elastic/ingest-docs/issues/1547

---

Screenshot 2024-12-16 at 1 26 26 PM

### DIFF
--- a/docs/en/ingest-management/elastic-agent/install-agent-msi.asciidoc
+++ b/docs/en/ingest-management/elastic-agent/install-agent-msi.asciidoc
@@ -10,11 +10,20 @@ The MSI package installer must be run by an administrator account. The installer
 
 . Download the latest Elastic Agent MSI binary from the link:https://www.elastic.co/downloads/elastic-agent[{agent} download page].
 
-. Run the installer:
+. Run the installer. The command varies slightly depending on whether you're using Bash or PowerShell.
++
+** Using Bash:
 +
 [source,shell]
 ----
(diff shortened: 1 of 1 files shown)


docs/en/ingest-management/elastic-agent/install-agent-msi.asciidoc
```diff
@@ -10,13 +10,22 @@ The MSI package installer must be run by an administrator account. The installer
...
 ----
 +
+** Using PowerShell:
++
+[source,shell]
+----
+./elastic-agent-<VERSION>-windows-x86_64.msi --% INSTALLARGS="--url=<URL> --enrollment-token=<TOKEN>"
+----
++
 Where:
```
shainaraskas, line 29: the picky person in me is worried this looks like it applies to just the default command prompt. consider pulling out the bullets and making them all child paragraphs.

shainaraskas (dismissed):
🏁 with a picky note

docs/en/ingest-management/elastic-agent/install-agent-msi.asciidoc
kilfoyle: @shainaraskas I've changed it to this, so the hierarchy is right. I'd like to avoid child paragraphs since it would differ from the formatting other, similar docs.

Here's how it looks now:

Screenshot 2024-12-16 at 3 21 14 PM

karenzone (approved):
LGTM 🚀

Issue #1618, opened 2025-01-10
[Fleet][Synthetic _source] Licensing comments and update to avoid deprecation warnings


lucabelluccini:
The page https://www.elastic.co/guide/en/fleet/current/data-streams-advanced-features.html must be updated to:
- Highlight that Synthetic _source requires an enterprise license starting 8.17.0 onwards
- Due to the deprecation https://github.com/elastic/elasticsearch/pull/116689, we must not tell users to use `"mappings._source.mode": "synthetic"` but instead use `"settings.index.mapping.source.mode": "synthetic"`
- ~We should tell that the procedure to disable Synthetic _source can work only in data streams which are not TSDS. I've tested it on 8.17 and it's impossible to disable Synthetic _source on a TSDS data stream.~
    - ~Afaik, across integrations, all data streams use TSDS and none use Synthetic _source explicitly.~
    - ~We might consider completely removing Synthetic _source guide.~

FYI @martijnvg (Elasticsearch) & @kilfoyle

martijnvg:
> We should tell that the procedure to disable Synthetic _source can work only in data streams which are not TSDS. I've tested it on 8.17 and it's impossible to disable Synthetic _source on a TSDS data stream.
Afaik, across integrations, all data streams use TSDS and none use Synthetic _source explicitly.
We might consider completely removing Synthetic _source guide.

@lucabelluccini How did you disable synthetic source in tsds data streams? You should be able to disable by configuring index.mapping.source.mode=stored.

lucabelluccini:
Hello @martijnvg 
I'm sorry but I didn't update the issue following the email thread we had internally.

Yes, it is possible to disable syn _source with TSDS. I was testing on an already existing index created prior 8.17.

I've updated the main issue text

kilfoyle:
Thanks @lucabelluccini! I've fixed up [the PR](https://github.com/elastic/ingest-docs/pull/1622) accordingly.

PR #1622, opened 2025-01-10
Add restrictions for synthetic `_source`

kilfoyle:
This adds the restrictions described in https://github.com/elastic/ingest-docs/issues/1618 to the [Enabling and disabling advanced indexing features for Fleet-managed data streams](https://www.elastic.co/guide/en/fleet/current/data-streams-advanced-features.html) page.

I also removed some extra spacing which I must have added earlier by accident.

@lucabelluccini and @martijnvg, please check that I have the `enable` and `disable` code examples right, since I'm kind of guessing. 

Closes: https://github.com/elastic/ingest-docs/issues/1618

**Previews:**

--- 

Screenshot 2025-01-21 at 9 40 23 AM

---

Screenshot 2025-01-21 at 9 40 03 AM

### DIFF
--- a/docs/en/ingest-management/data-streams.asciidoc
+++ b/docs/en/ingest-management/data-streams.asciidoc
@@ -704,7 +704,7 @@ Refer to <<data-streams-pipelines>> to learn more.
 * link:{ref}/mapping-source-field.html#synthetic-source[Synthetic `_source`]
 
 These features can be enabled and disabled for {fleet}-managed data streams by using the index template API and a few key settings.
-Note that in versions 8.17.0 and later, Synthetic `_source` requires an Enterprise license. As well, synthetic `_source` is supported only on non-TSDS data streams.
+Note that in versions 8.17.0 and later, Synthetic `_source` requires an Enterprise license. As well, disabling synthetic `_source` is supported only on non-TSDS data streams.
 
 NOTE: If you are already making use of `@custom` component templates for ingest or retention customization (as shown for example in <<data-streams-ilm-tutorial,Tutorial: Customize data retention policies>>), exercise care to ensure you don't overwrite your customizations when making these requests.
 
@@ -863,21 +863,8 @@ For example, the following payload disables TSDS on `nginx.stubstatus`:
 [[data-streams-advanced-synthetic-enable]]
 == Enable synthetic `_source`
 
-Note that synthetic `_source` is supported only on non-TSDS data streams.
-
 [source,json]
 ----
-PUT _component_template/<NAME>@custom
-{
-  "template": {
-    "mappings": {
-      "_source": {
-        "mode": "synthetic"
-      }
-    }
-  }
-}
-
 PUT _component_template/<NAME>@custom
 {
   "settings": {
@@ -897,7 +884,7 @@ PUT _component_template/<NAME>@custom
 [[data-streams-advanced-synthetic-disable]]
 == Disable synthetic `_source`
 
-Note that synthetic `_source` is supported only on non-TSDS data streams.
+Note that disabling synthetic `_source` is supported only on non-TSDS data streams.
 
 [source,json]
 ----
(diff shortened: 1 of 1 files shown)


kilfoyle:
@lucabelluccini, @martijnvg Please let me know if the warning looks okay.

Just trying to wrap things up since all docs PRs need to be merged by Jan.29. Thanks!

docs/en/ingest-management/data-streams.asciidoc
```diff
@@ -878,10 +863,11 @@ For example, the following payload disables TSDS on `nginx.stubstatus`:
 [[data-streams-advanced-synthetic-enable]]
 == Enable synthetic `_source`
 
+Note that synthetic `_source` is supported only on non-TSDS data streams.
```
martijnvg, line 866: Synthetic source is also supported on tsds data streams. I think this line can be removed?

docs/en/ingest-management/data-streams.asciidoc
```diff
@@ -891,20 +877,37 @@ PUT _component_template/<NAME>@custom
...
 PUT _component_template/<NAME>@custom
-
 {
-  "template": {
-    "mappings": {
-      "_source": {}
+  "settings": {
+    "index": {
+      "mapping": {
+        "source": {}
```
martijnvg, line 909:
```suggestion
        "source": {"mode": "stored"}
```

Otherwise it will fallback to the default. Which is synthetic source for logsbd and time_series index modes if license allows it.

martijnvg (commented):
I left two comments.
Additionally in the screenshot, I see the _source.mode mapping attribute being used. Can this be removed in favor for the index.mapping.source.mode index setting?

docs/en/ingest-management/data-streams.asciidoc
kilfoyle: Thanks! I've fixed it as you suggested.

docs/en/ingest-management/data-streams.asciidoc
kilfoyle: Thanks @martijnvg. I can remove the line, but from @lucabelluccini's [issue](https://github.com/elastic/ingest-docs/issues/1618): `I've tested it on 8.17 and it's impossible to disable Synthetic _source on a TSDS data stream.`

For now I've changed it to `Note that disabling synthetic _source is supported only on non-TSDS data streams.`
Please let me know if that's not what we want.

martijnvg: I commented on the issue: https://github.com/elastic/ingest-docs/issues/1618#issuecomment-2602660156

Let's see what went wrong here.

kilfoyle:
@martijnvg, @lucabelluccini  Based on discussion in https://github.com/elastic/ingest-docs/issues/1618#issuecomment-2604591612 I've removed the note about disabling Synthetic _source not being supported on on TSDS data streams.

docs/en/ingest-management/data-streams.asciidoc
```diff
@@ -720,6 +704,7 @@ Refer to <<data-streams-pipelines>> to learn more.
 * link:{ref}/mapping-source-field.html#synthetic-source[Synthetic `_source`]
 
 These features can be enabled and disabled for {fleet}-managed data streams by using the index template API and a few key settings.
+Note that in versions 8.17.0 and later, Synthetic `_source` requires an Enterprise license.
```
lucabelluccini, line 707: I don't know if you want so reuse the same wording as here:
https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-source-field.html#:~:text=To%20enable%20this%20subscription%20feature

kilfoyle: Thanks! I've added that wording, slightly rephrased: `Note that in versions 8.17.0 and later, synthetic `_source` is a subscription feature.`

I've also put the same note in the "Enable synthetic _source` section. The PR description has the latest screencaps.

lucabelluccini (approved):

karenzone (approved):
LGTM 🎸

PR #1627, opened 2025-01-20
Explain why Agent needs to connect to Kibana

nimarezainia:


docs/en/ingest-management/fleet/add-fleet-server-on-prem.asciidoc
```diff
@@ -98,6 +98,9 @@ You may need to allow access to these ports. Refer to the following table for de
 | {fleet-server} → {es} | 9200
 |===
 
+Note: Connectivity to {kib} on port 5601 is optional and not required at all times. {agent} and {fleet-server} may need to connect to {kib} if deployed in a
```
kilfoyle, line 101:
```suggestion
NOTE: Connectivity to {kib} on port 5601 is optional and not required at all times. {agent} and {fleet-server} may need to connect to {kib} if deployed in a 
```

Here's how it'll render:

Screenshot 2025-01-20 at 9 45 47 AM

PR #1722, opened 2025-03-06
Update fleet-agent-proxy-package-registry.asciidoc

lucabelluccini:
I've looked in the code and credentials can be provided via `xpack.fleet.registryProxyUrl: https://username:password@proxyhost:proxyport`. Note for security reasons this is only allowed when using `https`.
On top of this, I've discovered that when using HTTPS Proxy we also ignore invalid certs by default.

https://github.com/elastic/kibana/blob/2a32ed4755c0f4852dc234b72ce5ad278a0ad593/x-pack/platform/plugins/shared/fleet/server/services/epm/registry/proxy.ts#L40C9-L40C20

### DIFF
--- a/docs/en/ingest-management/fleet-agent-proxy-package-registry.asciidoc
+++ b/docs/en/ingest-management/fleet-agent-proxy-package-registry.asciidoc
@@ -17,7 +17,7 @@ example:
 xpack.fleet.registryProxyUrl: your-nat-gateway.corp.net
 ----
 
-If your HTTP Proxy requires authentication, you can include the 
+If your HTTP proxy requires authentication, you can include the 
 credentials in the URI, such as `https://username:password@your-nat-gateway.corp.net`, 
 only when using HTTPS.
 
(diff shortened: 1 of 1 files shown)


lucabelluccini:
This has to be back ported to at least 8.x and 7.17.x

lucabelluccini:
FYI @kilfoyle

kilfoyle:
Hi @lucabelluccini Thanks for opening this! The "Set the proxy URL" page exists only in the 8.x docs and not 7.17 (it was added in 8.10 I think). So I can backport into 8.18 and 8.17 (which is now current). 

Note that we're not allowed to update earlier, non-current versions of the docs, since we have a [banner](https://www.elastic.co/guide/en/fleet/8.16/epr-proxy-setting.html) at the top of those pages telling users that "No additional bug fixes or documentation updates will be released for this version.", and we don't want to false advertise. :-)

docs/en/ingest-management/fleet-agent-proxy-package-registry.asciidoc
```diff
@@ -17,6 +17,10 @@ example:
 xpack.fleet.registryProxyUrl: your-nat-gateway.corp.net
 ----
 
+If your HTTP Proxy requires authentication, you can include the
```
kilfoyle, line 20:
```suggestion
If your HTTP proxy requires authentication, you can include the 
```

kilfoyle (approved):

karenzone (approved):
LGTM

Issue #1757, opened 2025-04-11
[REQUEST]: Document Extended Key Usage for logstash output


jguay:
### Description

There is no mention of Extended Key Usage ([RFC 3280 extension point 4.2.1.13](https://www.ietf.org/rfc/rfc3280.txt)) required value(s) when certificates are not generated using [elasticsearch-certutil](https://www.elastic.co/guide/en/elasticsearch/reference/8.17/certutil.html) to [secure agent connection to fleet-managed logstash output](https://www.elastic.co/guide/en/fleet/current/secure-logstash-connections.html)

## Possible improvement

Confirm the requirements are same as for beats [example for filebeat doc](https://www.elastic.co/guide/en/beats/filebeat/8.17/configuring-ssl-logstash.html) and just copy the same statement at the beginning of https://www.elastic.co/guide/en/fleet/current/secure-logstash-connections.html#generate-logstash-certs: 

> If you choose not to use [certutil](https://www.elastic.co/guide/en/elasticsearch/reference/8.17/certutil.html), the certificates that you obtain must allow for both `clientAuth` and `serverAuth` if the extended key usage extension is present.

### Resources

Potentially check we use same libbeat code for logstash output managed by fleet so requirements documented for beats would apply equally

### Collaboration

The documentation team will investigate the issue and create the initial content.

### Point of contact.

**Main contact:**  @jguay 

**Stakeholders:**

kilfoyle:
@colleenmcginnis Thanks. I'm happy to hand it off. I don't think I can transfer authorship of the [PR](https://github.com/elastic/ingest-docs/pull/1758), but I don't mind keep it open if that helps. I'm reluctant to merge it until we have confirmation from engineering.

colleenmcginnis:
We can keep it open, and I'll just assign myself and shepherd it along if that's ok with you. 🙂

kilfoyle:
Sure thing. Thanks Colleen!

bmorelli25:
@karenzone I'm going to assign this to you for your next sprint. It looks like we just got the input we've been waiting on in @kilfoyle's PR https://github.com/elastic/ingest-docs/pull/1758.

PR #1758, opened 2025-04-11
Add note about extended key usage for Logstash output

kilfoyle:
This adds the note in green to the [Configure SSL/TLS for the Logstash output](https://www.elastic.co/guide/en/fleet/current/secure-logstash-connections.html#generate-logstash-certs) page.

screen

In the [docs issue](https://github.com/elastic/ingest-docs/issues/1757) @jguay asked: "Potentially check we use same libbeat code for logstash output managed by fleet so requirements documented for beats would apply equally"

@kpollich  Would you be able to help me with Julien's question, i.e, to verify if this requirement applies to the Fleet Logstash output as it does to Filebeat? 

Assuming that I should merge this PR, I'll then open another one for the 9.0+ docs.

### DIFF
--- a/docs/en/ingest-management/security/logstash-certificates.asciidoc
+++ b/docs/en/ingest-management/security/logstash-certificates.asciidoc
@@ -28,11 +28,15 @@ cluster. For more information, refer to the
 You can use whatever process you typically use to generate PEM-formatted
 certificates. The examples shown here use the `certutil` tool provided by {es}.
 
-TIP: The `certutil` tool is not available on {ecloud}, but you can still use it
+[TIP] 
+==== 
+* The `certutil` tool is not available on {ecloud}, but you can still use it
 to generate certificates for {agent} to {ls} connections. Just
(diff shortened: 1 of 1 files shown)


kilfoyle:
@kpollich Would any of the Fleet devs have some time to check on the requirement added here? It's way over my head. :-)

kpollich:
Hey @kilfoyle it looks like we missed this PR last year. @juliaElastic @criamico do you have any additional context that might help with getting this closed, or is this no longer relevant?

juliaElastic:
I don't know, this sounds more like a question to the Agent team cc @ebeahan @cmacknz

cmacknz:
The `ssl.*` options for all of our network connections are implemented by the same library (elastic-agent-libs) and behave identically with the same requirements for all of them (output, binary download, fleet server, etc).

kilfoyle:
Thanks for the info!

Closing this PR since I think the update is covered by @karenzone's https://github.com/elastic/docs-content/pull/5132

PR #1802, opened 2025-06-02
Allow backporting from branches other than main

colleenmcginnis:
Allows backporting from branches other than `main` so we can backport from `8.x` now that `main` is not being used for the Fleet and Agent guide. Prior art: https://github.com/elastic/observability-docs/pull/4109


kilfoyle (approved):
Cool!!! 🎸 
I never understood why I couldn't get auto-backporting to work off of 8.x.

